PR #17, opened 2019-10-22
Updating Travis to run chatbot tests as well as heartbeat tests

schwarrrtz:


schwarrrtz:
This seems to work. There is an error that comes up during some test runs on Travis that looks like this:

```UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'responder_phone_number' of null
    at sendButtonPressMessageForSession (/home/travis/build/bravetechnologycoop/BraveChatbot/server.js:114:46)```

However, it seems that this was present before we merged the chatbot and heartbeat repositories. The error copied above is actually from a Travis job testing the chatbot repo separately, but the same error sometimes occurs with the merged repo.

In any case, it seems like the goal of this feature branch has been achieved without introducing new bugs, so I'm going to merge this. Nice work @trisapeace :)

Issue #8, opened 2019-09-18
Add a test harness for the twilio functions


mariocimet:
We aren't testing the fallback phone number functionality with the current test harness, and doing so requires having more than the single test phone number that twilio provides us. The goal for this issue is to set up a small mocked version of the twilio API to be able to test this.

mariocimet:
Documenting a first try at this:

we're defining server.faketwilio, and assigning it a value within each test we want to use it in. Based on whether or not we're in the 'test' environment we use real or fake twilio.

mariocimet:
Right now the tests all pass whether or not twilioClient is actually defined. None of our tests depend on what the twilioClient returns (or whether it actually sends the message we want it to).

mariocimet:
In server.js, insofar as we care at all about what the twilio client returns its only to log it.

mariocimet:
What do we actually want to do with the responses from Twilio? If the API request fails?

@trisapeace says retry at least once, and possibly message somebody (Slack?) at Brave if that happens.

mariocimet:
Mocking right now is probably not the priority - I'll go ahead and try to write a test for the fallback phone which may or may not involve mocking.

schwarrrtz:
ok so @mariocimet wrote a test for the fallback alert that did not involve mocking. i'd say that's a good first step towards testing the twilio functionality. we can reassess whether we want to build out a test harness for the rest of the twilio code as a separate issue.

PR #21, opened 2019-11-08
Testing that a fallback alert is actually sent

mariocimet:
See issue #8 for discussion that led to this

### DIFF
--- a/chatbot/SessionState.js
+++ b/chatbot/SessionState.js
@@ -11,7 +11,7 @@ const incidentTypes = {
 
 class SessionState {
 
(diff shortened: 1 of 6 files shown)


mariocimet:
this moreso checks that a fallback alert has been actually *queued* - meaning the API call was at least accepted by twilio. This says nothing about the upstream carriers that twilio *itself* depends on to actually deliver the message. If we want to track the delivery of twilio messages more thoroughly we could make a new post request endpoint in our app for updates on the message status using their status callback url fields.

https://support.twilio.com/hc/en-us/articles/360008989454-Tracking-the-Delivery-Status-of-an-Outbound-Twilio-SMS-or-MMS-Message

schwarrrtz:
nice work!

PR #23, opened 2020-02-20
Systems on the heartbeat dashboard can be hidden

schwarrrtz:
This pull requests adds the ability for systems to be hidden from the heartbeat dashboard, so they don't clutter the view but the information about their system ID and name is saved for future log analysis.


mariocimet:
question - how do you usually interact with this API? do you use the utils.py file from the shell? should there be a new function there as well?

schwarrrtz:
I've been using Postman lately, but it would make sense to add a new function to utils.py too - I see you've done that in #24

schwarrrtz:
@mariocimet reports testing this and finding that everything works as expected. i've also tested it.

PR #25, opened 2020-02-26
Made a tool to help pair large numbers of buttons

schwarrrtz:
This tool runs on a Raspberry Pi and uses the Flic Linux SDK. It provides an interactive command-line process for pairing buttons and associating unit numbers and phone numbers with the UUID of each button. It outputs a CSV file in the format that the add_installation.sh script uses. Overall this greatly streamlines the process of preparing for a new installation.

The tool is tested up to the point of generating the CSV file, and seems to work well. The CSV files that the tool creates haven't been tested with the add_installation.sh script yet.

### DIFF
--- a/pairing-tool/.gitignore
+++ b/pairing-tool/.gitignore
@@ -1,0 +1,2 @@
+*.csv
+pairing_tool_sqlite.db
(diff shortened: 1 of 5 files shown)


schwarrrtz:
Code review attendees: Oona, Shawna, Sampath, Mario, Gordon, Dave

- Discussed the pairing tool as an initial test of the Flic Linux SDK, which could be used to replace the Flic Hub and further automate new installation setup. Discussed the limitations of this automation: buttons would still need to be manually paired with hubs, but this process would be much faster with a custom replacement for the Flic Hub.
- Discussed the testing of this pairing tool with the add_buttons_to_installation.sh script

PR #26, opened 2020-03-04
Added installation-specific chatbot response text

schwarrrtz:
There is now a system for changing the chatbot response text on a per-installation basis, using subclasses of SessionState. 

There are some basic tests for the new subclass, but this still needs to be tested in the context of a full system with real buttons etc.

### DIFF
--- a/chatbot/SessionState.js
+++ b/chatbot/SessionState.js
@@ -1,13 +1,5 @@
 const STATES = require('./SessionStateEnum.js');
 
-const incidentTypes = {
-    '0': 'Accidental',
-    '1': 'Safer Use',
-    '2': 'Unsafe Guest',
-    '3': 'Overdose',
-    '4': 'Other'
-};
-
 class SessionState {
 
     constructor(id, installationId, buttonId, unit, phoneNumber, state, numPresses, createdAt, updatedAt, incidentType, notes, fallBackAlertTwilioStatus) {
@@ -23,6 +15,14 @@ class SessionState {
         this.incidentType = incidentType
         this.notes = notes
         this.fallBackAlertTwilioStatus = fallBackAlertTwilioStatus
+        this.incidentCategoryMessage = 'Now that you have responded, please reply with the number that best describes the incident:\n0 - accidental\n1 - safer use\n2 - unsafe guest\n3 - overdose\n4 - other'
+        this.allowedIncidentTypes = {
+            '0': 'Accidental',
+            '1': 'Safer Use',
+            '2': 'Unsafe Guest',
+            '3': 'Overdose',
+            '4': 'Other'
+        }
     }
 
     advanceSession(messageText) {
@@ -32,11 +32,11 @@ class SessionState {
         switch (this.state) {
             case STATES.STARTED:
                 this.state = STATES.WAITING_FOR_CATEGORY;
-                returnMessage = 'Now that you have responded, please reply with the number that best describes the incident:\n0 - accidental\n1 - safer use\n2 - unsafe guest\n3 - overdose\n4 - other';
+                returnMessage = this.incidentCategoryMessage
                 break;
             case STATES.WAITING_FOR_REPLY:
                 this.state = STATES.WAITING_FOR_CATEGORY;
-                returnMessage = 'Now that you have responded, please reply with the number that best describes the incident:\n0 - accidental\n1 - safer use\n2 - unsafe guest\n3 - overdose\n4 - other';
+                returnMessage = this.incidentCategoryMessage
                 break;
             case STATES.WAITING_FOR_CATEGORY: {
                 let isValid = this.setIncidentType(messageText.trim());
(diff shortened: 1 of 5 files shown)


schwarrrtz:
I'm not sure what you mean - any implementation of this feature would require changes to both the code and the database, as this implementation does. If you're asking why the specific chatbot text is stored in code rather than the database, that's a fair question. I didn't see a need to store the chatbot text in the database because the number of installations we plan to support is fairly limited in the short to medium term, and because we plan to replace this chatbot implementation entirely with a mobile app. Also, further modifications to the chatbot text would not require nearly this many changes to the code, now that the subclass system is in place.

schwarrrtz:
Code review attendees: Oona, Shawna, Sampath, Gordon, Mario, Dave

- I will do a proper integration test with test buttons and a test phone. Assuming this goes well, I will merge this.
- note: ODetect has similar requirements to those that are addressed by this pull request

schwarrrtz:
@rhoberman, after thinking about this some more I have a better answer to your question. I think the underlying reason why I implemented this the way I did is because the state machine which implements the chatbot behaviour is based around the SessionState class. This worked well when we didn't have to support multiple installations, but now that there is more installation-specific data being used by the state machine, we need to either pass that data around as properties of the SessionState objects (which then means it makes sense to persist the data in the database 'sessions' table, which would be a ton of data duplication and would seem very weird when it should clearly be stored in the 'installations' table) or store the data in code.

schwarrrtz:
I think for the short term it makes sense to do it the way I did, but your question is making me think about tweaking the larger-scale structure of the codebase - maybe there should be a StateMachine class that is instantiated with installation-specific data, and then processes SessionState objects.

chatbot/SessionState.js
```diff
@@ -23,6 +15,14 @@ class SessionState {
...
+        this.allowedIncidentTypes = {
+            '0': 'Accidental',
+            '1': 'Safer Use',
+            '2': 'Unsafe Guest',
+            '3': 'Overdose',
+            '4': 'Other'
+        }
     }
 
     advanceSession(messageText) {
```
rhoberman: 1. Might be clearer to deal with the validity of the message upfront.

```
advanceSession(messageText) {
   ...
   case STATES.WAITING_FOR_CATEGORY:
      try {
         const incidentType = parseIncidentType(messageText)
         this.state = STATES.WAITING_FOR_DETAILS
         return 'Thank you. If you like, you can reply with any further details about the incident.'
      } catch (err) {
         return "'Sorry, the incident type wasn\'t recognized. Please try again."
      }
  ...
}

parseIncidentType(text) {
   // throw if unrecognised
}
```

2. Echo the invalid message to help the user diagnose the problem.

e.g. something like `Sorry, I didn't understand "accdental". Please try again. Valid options are Accidental, Safer Use, Unsafe Guest, Overdose, Other`

3. Default clause
Response is "Unfortunately, we have encountered an error in our system and will deal with it shortly.".

How would we know there has been an error? Recommend either explicitly doing whatever needs to happen to notify us here (e.g. log to Sentry) or propagating the error, e.g.

```
safeAdvanceSession(messageText) {
  try {
    const responseText = advanceSession(messageText)
    safeSendResponse(responseText)
  } catch (error) {
     safeSendResponse("Unfortunately, we have encountered an error in our system and will deal with it shortly.")
     throw error
  }

 // caught and reported by top-level Sentry handler
```

or something to that effect

chatbot/db/db.js
```diff
@@ -11,7 +11,19 @@ pool.on('error', (err) => {
     console.error('unexpected database error:', err)
 })
 
-function createSessionFromRow(r) {
+async function createSessionFromRow(client, r) {
+
+    if(!r) {
+        return null
```
rhoberman: Is there any valid case where this would be invoked with a null row? If not, recommend throwing an error as this graceful handling will make it much harder to detect and understand a problem higher-up the chain.

chatbot/db/db.js
```diff
@@ -11,7 +11,19 @@ pool.on('error', (err) => {
...
+    }
+
+    let { rows } = await client.query("SELECT * FROM installations WHERE id = $1", [r.installation_id])
+    let className = rows[0].session_state_class_name
```
rhoberman: throwing an "installation not found (id=x)" error would make it easier to diagnose the cause if this receives an invalid installation.

chatbot/db/db.js
```diff
@@ -41,14 +53,17 @@ module.exports.getUnrespondedSessionWithButtonId = async function(buttonId, clie
     const query = "SELECT * FROM sessions WHERE button_id = $1 AND state != $2 AND state != $3 AND state != $4"
     const values = [buttonId, STATES.WAITING_FOR_CATEGORY, STATES.WAITING_FOR_DETAILS, STATES.COMPLETED]
     const { rows } = await client.query(query, values)
-   
+
+    let session = await createSessionFromRow(client, rows[0])
+
     if(!transactionMode) {
```
rhoberman: Does this need to be in a finally block? i.e. would the client be released if an error occured during this method?

schwarrrtz:
We've merged an alternate implementation of this pull request: https://github.com/bravetechnologycoop/BraveButtons/pull/33

PR #28, opened 2020-03-12
restart heartbeat & chatbot servers once per week

schwarrrtz:
this will hopefully address the certificate renewal bug once and for all.

### DIFF
--- a/chatbot/setup_server.sh
+++ b/chatbot/setup_server.sh
@@ -51,7 +51,10 @@ else
     certbot certonly --standalone 
 
     echo "0 0 * * 0 certbot renew --pre-hook 'env HOME=$HOME pm2 stop server' --post-hook 'env HOME=$HOME pm2 start server'" > crontab.tmp
+
+    # reboot weekly to ensure that renewed certificates are loaded in memory by the server
     echo "0 0 * * 1 env HOME=$HOME pm2 restart server" >> crontab.tmp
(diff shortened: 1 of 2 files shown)


schwarrrtz:
this is currently being tested on chatbot-dev.brave.coop. it should restart on Monday afternoon.

schwarrrtz:
Good point, I've added comments. The choice of a weekly reboot interval was somewhat arbitrary. It doesn't need to be daily since certificate renewal happens on a monthly (or longer) time scale.

Whether or not a load balancer (managed with Kubernetes or otherwise) would remove the need for us to manage certificates ourselves is a topic for another discussion. 😃

schwarrrtz:
Action items from code review:

- determine when the certificate will be renewed next & create a Clickup task to double check that the server has restarted properly
- consider adding Sentry to monitor server restarts

schwarrrtz:
So, interestingly, this didn't work in the test. On chatbot-dev.brave.coop, pm2 is reporting 28 days of uptime for the server process. So clearly there was no restart on Monday, as expected.

schwarrrtz:
I think I've fixed the underlying issue stopping the server from restarting after a certificate renewal:

https://github.com/bravetechnologycoop/BraveButtons/pull/31

Since this pull request should no longer be necessary, I'm closing it.

PR #31, opened 2020-03-23
fixed an issue preventing server restart after certificate renewal

schwarrrtz:
Cron tasks execute with a minimal environment, meaning that the location of the pm2 executable was not on the PATH used by cron. This prevented the server process from restarting after certificate renewal took place.

I didn't catch this issue while testing the previous certificate renewal fix (#20) because I was copying the command from the crontab and executing it from a root shell, which had a different PATH setting. I tested this latest fix by changing the crontab to execute the command every minute instead of every week. In order to view the output of the cron jobs, I had to install a mail transfer agent (I used postfix, configured for local mail delivery only).

Relevant info:

https://askubuntu.com/questions/222512/cron-info-no-mta-installed-discarding-output-error-in-the-syslog
https://community.letsencrypt.org/t/problem-with-renew-certificates-the-request-message-was-malformed-method-not-allowed/107889/22

### DIFF
--- a/chatbot/setup_server.sh
+++ b/chatbot/setup_server.sh
@@ -49,12 +49,16 @@ else
     ./setup_postgresql.sh $installationName $responderNumber $fallbackNumber
 
     certbot certonly --standalone 
-    
-    echo "PATH=/bin:/usr/bin:/usr/local/bin" > crontab.tmp
-    echo "0 0 * * 0 certbot renew --pre-hook 'env HOME=$HOME pm2 stop server' --post-hook 'env HOME=$HOME pm2 start server'" >> crontab.tmp
-    crontab crontab.tmp
-    rm crontab.tmp
-    
+
+    mkdir -p /var/log/brave
+    touch /var/log/brave/cron.log
+
+    # set up cron to have certbot renew certificates regularly, then send output to a log file
+    echo "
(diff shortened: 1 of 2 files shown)


schwarrrtz:
So cron requires that a mail transfer agent is installed in order to route the output from the cron jobs somewhere. As far as I can tell there is no other option to be able to view the output of the cron jobs. This seems to be some very old-school functionality that never got modernized.

However, it's not the same thing as email. Installing postfix and configuring for local mail delivery essentially turns it into an overly complicating logging tool - the 'mail' just gets written into a file in `/var/mail` depending on which user the mail is addressed to. In this case, since all the cron jobs are added to root's crontab, the output of the cron jobs gets written to `/var/mail/root` which is just a regular text file.

schwarrrtz:
It turns out that installing postfix interactively is a huge pain, and it's much easier just to pipe the output from the cron job to a log file from within the cron job itself. I'm going to add that code to this pull request.

schwarrrtz:
The new changes to logging and the install scripts have been tested on chatbot-dev.brave.coop and seem to work well.

schwarrrtz:
Code review attendees: Dave, Sampath, Mario

Consensus: good to go!

PR #32, opened 2020-03-23
fixed rfkill issue affecting Raspbian Buster

schwarrrtz:


schwarrrtz:
Code review attendees: Sampath, Mario, Dave

Testing protocol (TODO): run setup script without modification, confirm that it is broken. Run setup script again with only this modification, confirm that it works.

PR #37, opened 2020-07-29
Fix Travis

trisapeace:
Travis automatically updated Node to version 14. This caused the tests to stop working about 4 months ago. 

We now specify the latest LTS version, which is Node v12.
We also noticed that the version of Postgres that Travis was using was 9.6, but we use 9.5 on production so we changed to use that now instead to better match production.


schwarrrtz:
Theresa and I both feel confident that this fixes the Travis issue. We're going to merge it now.

PR #39, opened 2020-08-13
Hardcode chatbot's and heartbeat's node version to 12.18.3 (CU-acgyqr)

trisapeace:
ClickUp: https://app.clickup.com/t/acgyqr

I checked on `chatbot.brave.coop` and it is currently using version `12.16.1`.

I think that we should take this opportunity to upgrade to `12.18.3` because this is the current `stable` version. So I did that in the `setup_server.sh` script as well as for Travis. It seems like a low-risk change because it's only a minor version change. The tests pass locally and on Travis with that node version. I imagine we'll need to run this script again on the servers after this merge to make this change effective, right?

Using `node_js: 12` in the `.travis.yml` file only allows us to specify the major version. To get more fine-grained control, it uses a `.nvmrc` file (https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions-using-nvmrc). Which may be confusing since we're not using NVM, but hopefully not too bad.

I also changed `agt-get` to `apt` while I was in that part of the code.


trisapeace:
Testing if this will connect to ClickUp: CU-acgyqr

trisapeace:
Task linked: [CU-acgyqr specify node 12 in chatbot install script](https://app.clickup.com/t/acgyqr)

schwarrrtz:
just noting here that `.nvmrc` will also apply to the heartbeat server, which is also a node app. maybe we should also update `heartbeat/setup_server.sh` to keep everything consistent?

also, this is yet another indicator that we're duplicating work between the chatbot and heartbeat servers... but that's another issue.

trisapeace:
Agreed, I think doing the heartbeat server at this time would also be a good idea to keep them consistent and pinned until we have time to combine them into one server. I'll do that.

As for the `.nvmrc`. It is only being used by Travis (since we use `n` in the setup scripts) and the heartbeat isn't running tests right now so it shouldn't actually affect it right now. Speaking of which, do we have a ClickUp task to add tests for this?

chatbot/setup_server.sh
```diff
@@ -33,13 +33,13 @@ else
...
+    apt install software-properties-common
     add-apt-repository -y ppa:certbot/certbot
-    apt-get update
-    apt-get install -y nodejs npm certbot postgresql postgresql-contrib
+    apt update
+    apt install -y nodejs npm certbot postgresql postgresql-contrib
     npm install -g pm2 n
-    n stable
+    n 12.18.3         # keep this in sync with .nvmrc for Travis and with heartbeat/setup_server.sh
     npm install
```
trisapeace, line 43: @schwarrrtz Just wondering if there's a reason that the `npm install` command is in the different place in these two install scripts?

trisapeace: I'll move `npm install` in the other file to the bottom of this section to match the chatbot.

schwarrrtz: don't think there's a reason.

schwarrrtz:
looks good to me!

PR #41, opened 2020-08-20
Add CHANGELOG.md (CU-5wd4g9)

trisapeace:
A changelog following the structure described in https://keepachangelog.com/en/1.0.0/ 

I didn't go into any detail of the first versions (`1.0-pi-heartbeat` and `1.0-chatbot`). My thinking was that it's unlikely that anyone will need to look back that far and if I were to fill it in properly it would just be an exercise in describing an out-of-date system.

For the release dates, I just used the date of the commit with the tag on it. I'm pretty sure that this is often not the release date and/or the code was released on different devices on different days. Another option would be to just leave the dates off if I'm unsure about when they were actually released. I'm leaning towards showing this inaccurate dates because at least they give us some idea (hopefully nothing was released *before* these dates).

Please let me know what you think about these decisions and about the changelog in general. It would be great if we could standardize on this across our repos.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,134 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+Please note that the date associated with a release is the date the code
+was committed to the `master` branch. This is not necessarily the date that
+the code was deployed.
+
+## [Unreleased]
+### Added
+- Changelog.
+
+## [1.4.3] - 2020-08-18
+### Added
+- Installation instructions to the README.
+
+### Changed
+- Replace cron cert renewal with simple weekly restart.
+- Update node to v12.18.3 for heartbeat and chatbot.
+- Installation script improvements.
+
+### Fixed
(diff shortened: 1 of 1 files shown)


trisapeace:
Task linked: [CU-5wd4g9 add changelog to BraveButtons repo](https://app.clickup.com/t/5wd4g9)

sampathsatti:
I would advocate having the dates left in even though they're inaccurate. I'd suggest prefacing the changelog with that rider, but I think the dates themselves tell a nice (development) story. Can we also explicitly call out that they are YYYY-MM-DD?

The changelog format is great, and doesn't seem like too much overhead whenever a new version is released.

I'd say merge this and this is a great point to bring up keepachangelog's comment on rewriting changelogs using pull requests being acceptable - In the ICBD framework, this is definitely safe to try...

trisapeace:
Thanks for taking a look @sampathsatti . I like your idea of adding a note about the dates being the date the changes were committed into `master` and not necessarily the deployment date of that code.

As for the YYYY-MM-DD, I feel like we shouldn't need to call this out because it is specified as such in keepachangelog

schwarrrtz:
this is great. I really like the 'unreleased' section as a way of tracking changes in development - this makes reasoning about the effect a release will have much easier :)

mariocimet:
I'm down w/ this!

PR #40, opened 2020-08-15
Changes allow Buttons to use a remote PostgreSQL service (CU-5r8y5w + CU-5jexvw)

mariocimet:
- updated .env schema to reflect need to specify a remote host and port for the managed db
- updated setup scripts to use new values. In production, role and database creation will be done via the Managed DB GUI, keeping those commands in scripts for use in travis and local installs
- updated db.js with parameters for connecting to remote DB
- updated travis.yml to use PostgreSQL 12, which is also the version that runs on the managed DB on DigitalOcean

On the DigitalOcean side you manage this from [https://cloud.digitalocean.com/databases/button-db](https://cloud.digitalocean.com/databases/button-db), including creating users, databases, and setting up connection pools

Database migration can be done by dumping an existing database with 
`pgdump --data-only -O pg_brave > bravedata.sql` 
and feeding the resulting dump file to psql with 
`PGPASSWORD=password psql -U db_user -h dbhost.com -p 12345 -d targetdatabase --set=sslmode=require < bravedata.sql`. I've tested this and the setup script changes on [chatbot-dev-2.brave.coop
](https://chatbot-dev-2.brave.coop/)

Screenshot from 2020-08-14 17-17-20

PR #42, opened 2020-08-24
Add unmute API endpoint to Heartbeat (CU-baj2pv)

trisapeace:
Added a new endpoint called `POST unmute_system` that will unmute a muted system.

To help with testing this feature (and because I think it will be useful in real life), I also added a new "Muted?" column to the Heartbeat Dashboard that will display "Y" for systems that are muted and "N" for systems that are not muted.
image

While I was in there, I also noticed that there is a `POST hide_system` endpoint but no `POST show_system`. I'm wondering if this should be added as well. What do you think @schwarrrtz ?

### DIFF
--- a/heartbeat/utils.py
+++ b/heartbeat/utils.py
@@ -23,6 +23,24 @@ def send_rename_request(server_url, system_id, name):
         print("error sending rename request")
         print(e)
 
+def send_mute_request(server_url, system_id):
+    payload = {"system_id": system_id}
+    try:
+        r = requests.post(server_url + r"/mute_system", json=payload)
+        print("response to mute request:", r.status_code, r.reason)
+    except Exception as e:
+        print("error sending mute request")
+        print(e)
+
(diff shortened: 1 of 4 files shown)


trisapeace:
Task linked: [CU-baj2pv Provide a way to un-mute heartbeat alerts through the API](https://app.clickup.com/t/baj2pv)

heartbeat/utils.py
```diff
@@ -23,6 +23,24 @@ def send_rename_request(server_url, system_id, name):
         print("error sending rename request")
         print(e)
 
+def send_mute_request(server_url, system_id):
```
trisapeace, line 26: I just added those utility functions, @schwarrrtz .

schwarrrtz:
This looks good to me - we're going to hold off merging this until Mario merges #40 to avoid excessive rebasing.

schwarrrtz (approved):

mariocimet (approved):

PR #44, opened 2020-08-28
Add logging to help debugging (CU-behg93)

trisapeace:
- Seeing behaviour where Last Seen (Ping) is larger than Last Seen
(Flic), which doesn't make a lot of sense. So trying to see
exactly which IP address is being pinged, what the return code
from ping was, and which darkstat line provided the IP
- Also now log the FlicNotFoundError message from the object rather
than a generic message
- Added corresponding tests

I couldn't figure out how to test this on my local RPi because every time I plugged it in, I took down our local network. My guess is that because it was searching for the non-existent FlicHub every second, it was bogging down the system. Instead, I made mini python scripts to test just the functions that I changed locally to see that the logs turned out the way that I expected, and I added unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ the code was deployed.
 ### Added
 - Changelog (CU-5wd4g9).
 - Environment variables to Travis config (CU-b4m32r).
+- More logging to the Raspberry Pi (CU-behg93).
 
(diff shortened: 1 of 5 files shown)


trisapeace:
Task linked: [CU-behg93 Investigate high Last Seen (Ping) for Buchan hubs](https://app.clickup.com/t/behg93)

PR #45, opened 2020-08-29
Use .env variable for fallback "from" phone number instead of the phone number of the room whose button press was unanswered (CU-6ed85y)

trisapeace:
- To ensure that these messages are not confusing for the fallback
staff, added the installation name to the message
- Unfortunately, the phone number that we are using in the tests for
the test rooms and for the test fallback from number need to be the
same because there is only one number that Twilio provides that will
succeed as a From number while using the test credentials. So I wasn't
able to write a test for this change

I pulled this change onto `chatbot-dev.brave.coop` and tested with my own phone as the responder and fallback phone. It worked as I expected.


trisapeace:
Task linked: [CU-6ed85y change fallback alert sender number](https://app.clickup.com/t/6ed85y)

PR #49, opened 2020-09-14
CU-byp1m3 + CU-3j6jj7 - Investigate Flakey Test

mariocimet:
There's some documentation about how I investigated this in the clickup task, and test results for batch runs of tests. The only change here is to increase the amount of time that the test waits in between sending a test button press and checking the resulting session. This gives more time for the reminder and fallback message timers to expire, and for the several asynchronous calls that need to happen in between that and the message actually going out to conclude.

The timeouts add to 3000 milliseconds, so increasing from 3500 to 4000 doubles the extra time for those asynchronous calls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ the code was deployed.
 - Upgrade handlebars (CU-c6rgqh).
 - Upgrade yargs-parser (CU-c6rgqh).
 
+### Changed
+- db.js now loads different environment variables depending on NODE_ENV test flag
+- Increased wait time in fallback message test to address race condition and reduce test flakiness
+
 ## [1.5.0] - 2020-09-04
 ### Added
 - Changelog (CU-5wd4g9).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,8 @@ the code was deployed.
 - Upgrade yargs-parser (CU-c6rgqh).
 
 ### Changed
-- db.js now loads different environment variables depending on NODE_ENV test flag
-- Increased wait time in fallback message test to address race condition and reduce test flakiness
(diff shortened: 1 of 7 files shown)


trisapeace:
Task linked: [CU-3j6jj7 look into flaky test: 'should send a message to the fallback phone number if enough time has passed without a response'](https://app.clickup.com/t/3j6jj7)

trisapeace:
Task linked: [CU-byp1m3 add test versions of PG_HOST and PG_PORT .env fields](https://app.clickup.com/t/byp1m3)

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ the code was deployed.
 - Upgrade handlebars (CU-c6rgqh).
 - Upgrade yargs-parser (CU-c6rgqh).
 
+### Changed
+- db.js now loads different environment variables depending on NODE_ENV test flag
```
trisapeace, line 17: Could you please add the ClickUp IDs to these items?

chatbot/db/db.js
```diff
@@ -5,18 +5,23 @@ const { Pool } = require('pg')
...
+    database: getEnvVar('PG_USER'),
+    password: getEnvVar('PG_PASSWORD'),
     ssl: true
 })
 
 pool.on('error', (err) => {
     console.error('unexpected database error:', err)
 })
 
+function getEnvVar(name) {
```
trisapeace, line 20: This is just copy-and-pasted from `server.js`, right?

Instead, you could create new `helpers.js` file with this function in it and the use it in both `server.js` and here in `db.js`. That way if we ever need to change it for some reason, it'll work in both places.

trisapeace (commented):
Thanks so much for investigating this!

trisapeace (approved):
Looks good. Thanks!

schwarrrtz:
looks good.

PR #52, opened 2020-09-30
Add an heartbeat alert text message for exceeding a ping threshold (CU-crut3u)

trisapeace:


trisapeace:
Task linked: [CU-crut3u Add heartbeat alert when Last Seen (Ping) is above a threshold](https://app.clickup.com/t/crut3u)

heartbeat/server.js
```diff
@@ -26,15 +26,16 @@ db.on('compaction.done', () => {
 
 const FLIC_THRESHOLD_MILLIS = 180*1000
 const HEARTBEAT_THRESHOLD_MILLIS = 60*1000
+const PING_THRESHOLD_MILLIS = 320*1000
```
trisapeace, line 29: What do you think about a 3-minute threshold for this.

I went back through the logs since Mario changed the wifi config and I do see periods where the Ping was unsuccessful for up to 389 seconds and then recovers on its own. It looks like most of the time it recovers in less than 60 seconds, but there are still quite a few times when it's up in the 100 or 200 second range before it self-recovers.

I want to balance the terrible-ness of it being down vs how long it would take one of us to get to a computer to manually fix it in any case.

On the other hand, knowing that these numbers can be so high, maybe we want to go back to the idea of the brute-force auto-restart after a much lower threshold that would take the system down but would at least show the red light to the Resident to let them know that it didn't work.

heartbeat/server.js
```diff
@@ -175,24 +176,34 @@ function checkHeartbeat() {
             let flicDelayMillis = currentTime.diff(flicLastSeenTime)
             let heartbeatLastSeenTime = moment(doc.heartbeat_last_seen_time)
             let heartbeatDelayMillis = currentTime.diff(heartbeatLastSeenTime)
+            let pingLastSeenTime = moment(doc.flic_last_ping_time)
```
trisapeace, line 179: I noticed that `flic_last_ping_time` should be `ping_last_seen_time` to match the other two, however it looks like it's already being used in the "DB".

@mariocimet does the Heartbeat use Postgres in the combined Buttons server?

heartbeat/server.js
```diff
@@ -26,15 +26,16 @@ db.on('compaction.done', () => {
...
 function log(logString) {
     console.log(moment().toISOString() + " - " + logString)
 }
 
-function sendAlerts(systemName) {
+function sendAlerts(connectionType, systemName) {
     for(let i=0; i<config.TWILIO_TO_NUMBERS.length; i++) {
         twilioClient.messages.create({
-            body: `The Flic connection for ${systemName} has been lost.`,
+            body: `The ${connectionType} connection for ${systemName} has been lost.`,
```
trisapeace, line 38: Is it useful to know which threshold triggered the heartbeat alert?

heartbeat/server.js
mariocimet: I think another thing to consider is that a high threshold like 6 minutes in this case makes it likelier that we don't tune out the alerts and would probably signal truly exceptional/prolonged downtime cases?

Then there's the separate issue of tracking and working on eliminating the sub-threshold outages? I don't know if humans intervening on a command line somewhere are the right solution to those

heartbeat/server.js
mariocimet: it does yes, these NeDB calls all get switched

heartbeat/server.js
mariocimet: I think so! especially if the response to this is a human hopping on the command line - and starting to debug

mariocimet:
This looks good and straightforward! Any thoughts on what order to do this in re: the merging servers changes?

heartbeat/server.js
schwarrrtz: I'm not convinced that these short-duration episodes where the hub doesn't respond to pings actually represent real downtime. I'd want to see evidence of that before we start aggressively restarting the flic hub, especially since that would introduce a known amount of downtime while the flic hub reboots.

I think if it's commonly reaching ~200 seconds and then recovering on its own we should set the threshold so that these events don't trigger alerts. I see that you've got it set at 320 seconds or ~5 mins, which seems appropriate.

heartbeat/server.js
schwarrrtz: I agree, `ping_last_seen_time` makes more sense.

heartbeat/server.js
schwarrrtz: Yes! This is great.

heartbeat/server.js
```diff
@@ -175,24 +176,34 @@ function checkHeartbeat() {
...
+            let pingDelayMillis = currentTime.diff(pingLastSeenTime)
         
             if(flicDelayMillis > FLIC_THRESHOLD_MILLIS && !doc.sent_alerts) {
                 log(`flic threshold exceeded; flic delay is ${flicDelayMillis} ms. sending alerts for ${doc.system_name}`)
                 updateSentAlerts(doc.system_id, true)
                 if(doc.muted) {
                     return
                 }
-                sendAlerts(doc.system_name)
+                sendAlerts('Flic', doc.system_name)
```
schwarrrtz, line 188: This is ambiguous - can you change this to 'Darkstat' or 'Network monitor' or something?

heartbeat/server.js
trisapeace: That's a good point about us not knowing whether the short-duration episodes cause a real outage. We know that the long ones did.

Unfortunately to try it out we would need to be really on the ball and hope that someone at the Buchan is able to push a test button before it resolves itself. And even then, if the button worked and the last seen ping resolved itself at the same time, I wouldn't trust that we weren't just "lucky" that it resolved itself just when we wanted to press it.

As for tracking, these logs help but they aren't easy to parse out the valuable information. We could start logging to the DB now that we have one. But that probably isn't worth it unless we rule out the possibility of using the Flic Hub SDK.

So what I'm hearing from you both is that this looks fine as is for the time being. I think so too.

heartbeat/server.js
trisapeace: :+1: So this will be handled in PR #53

trisapeace:
@mariocimet . I'm open for suggestions for the merge with your code. Originally I was hoping to sneak this in before, but now they are moving along almost simultaneously. I think it would be easier for you to merge first. My changes are all in just one file (+ the CHANGELOG).

heartbeat/server.js
trisapeace: Done.

heartbeat/server.js
schwarrrtz: It occurs to me that `connectionType` is not really an accurate description of what the various metrics are. Similarly, something like `The Darkstat connection for BraveOffice has been lost.` is a bit misleading - what we actually want to say is more like `According to Darkstat data, the connection to BraveOffice has been lost.`

schwarrrtz: Because in this example there is probably nothing wrong with Darkstat - the problem is actually most likely with the Flic hub.

trisapeace: Good point. I've updated all the alert texts to make them more accurate. What do you about them now @schwarrrtz ?

trisapeace:
I just loaded this onto the Chatbot Dev server to try it out. All the different alerts work as I expected.

trisapeace:
@mariocimet @schwarrrtz 
I just added those small increases to the thresholds. Is this OK to merge into Mario's branch?

mariocimet:
I am in favour of merging this into the #53  branch

heartbeat/server.js
schwarrrtz: They all look good, except for the one typo I noted in another comment.

server/server.js
```diff
@@ -476,43 +477,61 @@ async function checkHeartbeat() {
...
+        let pingDelayMillis = currentTime.diff(pingLastSeenTime)
         
         if(flicDelayMillis > FLIC_THRESHOLD_MILLIS && !hub.sentAlerts) {
             log(`flic threshold exceeded; flic delay is ${flicDelayMillis} ms. sending alerts for ${hub.systemName}`)
             await updateSentAlerts(hub, 'true')
             if(hub.muted) {
                 continue
             }
-            sendAlerts(hub.systemName, helpers.getEnvVar('TWILIO_HEARTBEAT_FROM_NUMBER'), hub.heartbeatAlertRecipients)
+            sendAlerts('Darkstat has lots visibility of the Hub', hub.systemName, hub.heartbeatAlertRecipients)
```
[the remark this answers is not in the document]
trisapeace: Good catch :-)

heartbeat/server.js
trisapeace: Thanks! All fixed.

PR #55, opened 2020-10-08
Get Travis to lint the Server and the Pairing Tool (CU-cchv7x)

trisapeace:
- Moved the ESLint configuration file to the root so that any changes in it will automatically be used by both the pairing tool and the server
- Configured ESLint to ignore the `no-constant-condition` rule because it fails on `while(true)` which is used by the Pairing Tool.
- Added the pairing tool to Travis and configure it to run the linter
- Remove ESLint from the server dependencies because now using `npx` to download the latest every time
- `while(true)` fails the `no-constant-condition` recommended eslint rule, so needed to change that to `for(;;)` in the Pairing Tool
- Removed unused variables in the Pairing Tool because they violated lint rules
- Updated the README accordingly


trisapeace:
Task linked: [CU-bgke33 Combine chatbot and heartbeat into a single Button node app](https://app.clickup.com/t/bgke33)

trisapeace:
Task linked: [CU-cchv7x Use Travis to run ESLint in buttons server, and pairing tool](https://app.clickup.com/t/cchv7x)

pairing-tool/index.js
```diff
@@ -13,12 +13,12 @@ csvWriter.pipe(fs.createWriteStream(outputFile))
 
 async function scanAndPairOneButton() {
 
-    let promise = new Promise((resolve, reject) => {
+    let promise = new Promise((resolve, reject) => {    // eslint-disable-line no-unused-vars
```
trisapeace, line 16: @schwarrrtz 
What do you think about the disabling eslint inline for these function parameters that you wanted to keep?

schwarrrtz: this is a good compromise - I've tweaked these config comments in a commit that I just pushed to this branch.

server/package.json
```diff
@@ -22,11 +22,11 @@
     "twilio": "^3.33.2"
   },
   "devDependencies": {
-    "eslint": "^6.7.2",
```
schwarrrtz, line 25: is there a reason why you took this out? I prefer to have ESLint installed locally, rather than having `npx` pull down an entire install each time I want to run the linter.

trisapeace: Just so there is one fewer thing to keep up-to-date. They seem to release quite frequently (https://www.npmjs.com/package/eslint).

I'm OK with switching it back, but in that case we'll want to switch it in the `npm run lint` command as well to ensure that we're running it the same way in Travis. I just made a commit for that and you can let me know what you think.

pairing-tool/index.js
```diff
@@ -67,6 +71,7 @@ async function runPairingTool() {
 
     console.log('Welcome to the Brave Button pairing tool.')
 
+    // eslint-disable-next-line no-constant-condition -- we want to implement a run loop here
```
trisapeace, line 74: What do you mean by a run loop, @schwarrrtz ?
`for(;;)` should have the same behaviour as `while(true)`.

schwarrrtz: A run loop is a concept from microcontroller programming which is basically just `while(true) { do stuff } `

I'm not familiar with `for(;;)` - I wasn't aware that it has the same behaviour as `while(true)`. I feel like `while(true)` is more clear. I mentioned this in passing to Richard and he seemed to agree with me.

server/package.json
schwarrrtz: Ah yeah, fair. Maybe this would be an opportunity to try a more liberal semver? Or maybe we can leave that until our semver review of `package.json`

pairing-tool/index.js
trisapeace: I'm cool with that. Thanks for clarifying.

server/package.json
trisapeace: Let's leave it for the other task for now.

schwarrrtz:
this looks good to go.

PR #60, opened 2020-10-29
script for adding hubs from a csv file

mariocimet:


schwarrrtz:
looks good - you've tested this and made some small changes in response to my review.

schwarrrtz (approved):

server/db/add_hubs.sh
```diff
@@ -0,0 +1,50 @@
...
+            PG_HOST="$value"
+        elif [[ "$name" == "PG_PORT" ]]; then
+            PG_PORT="$value"
+        fi
+    done < $1
+
+
+    while IFS=",", read -r system_id system_name heartbeat_alert_recipients; do
+        if [[ "$system_id" != "system_id" && "$system_id" != "" ]]; then
+            echo "Adding button"
```
trisapeace, line 40: Do you mean `hub` instead of `button` here?

server/db/brave_hubs.csv.example
```diff
@@ -0,0 +1,3 @@
+system_id,system_name,heartbeat_alert_recipients
```
trisapeace, line 1: It would be nice if this file name matched the script name. So be `add_hubs.csv.example`. That way it would be nice and easy to find because they will be beside each other in the directory structure. Might also be worth mentioning in the script's help text that this example file exists.

trisapeace (commented):
Neither of my comments affects the running of this, but maybe they are some things that you'd want to change. Up to you.

server/db/add_hubs.sh
```diff
@@ -0,0 +1,50 @@
...
+    exit 1
+elif [[ ! -n "$2" ]]; then
+    echo ""
+    echo "Usage: $0 path_to_.env_file path_to_hubs_csv"
+    echo "" 
+    echo "Example: $0 ./../.env ./brave_hubs.csv.example"
+    echo ""
+    echo "The hubs CSV file"
+    echo "MUST have the header 'system_id, system_name, heartbeat_alert_recipients'"
+    echo "MUST use Unix line endings (LF)"
```
trisapeace, line 19: You have a copy of this file on `chatbot-dev` that is slightly different for this help text:
```
    echo "The registry CSV file"
    echo "MUST have the header 'system_id, system_name, heartbeat_alert_recipients'"
    echo "MUST use Unix line endings (LF), or else the phone numbers will have '\r' at the end"
    echo "MUST end with a newline, or else the last button will be silently ignored"
```

I've reverted it on `chatbot-dev`, but have recorded it here in case you wanted to include it in another commit.

server/db/add_hubs.sh
schwarrrtz: this is captured in https://app.clickup.com/t/faxfbg

server/db/add_hubs.sh
schwarrrtz: the version you're sharing here @trisapeace looks like it's just a more out-of-date version, ie. closer to the original `add_buttons_to_installation.sh` script. however, i did notice a few issues with this text while preparing to run the script during the leadup to the princess rooms installation. these issues are documented in https://app.clickup.com/t/faxfbg

server/db/brave_hubs.csv.example
schwarrrtz: i agree with this

server/db/add_hubs.sh
trisapeace: That makes sense. @mariocimet , can we delete the version that is chillin' on `chatbot-dev`?

PR #61, opened 2020-10-30
A number of tweaks to make the button-press handling more consistent (CU-f0w9qm)

trisapeace:
It might be easiest to go through this PR by going through the commits rather than looking at all the code at once.

There are four main changes:

1. When there is a **double click**, Power Automate / Flow sends the parameter `Type=double click`. However, in the code, we were looking for `double_click` (with an underscore). This means that when we received a request with `Type=double click`, we were treating it the same as a single click. This means that we might have missed the initial message for some urgent requests (they would still have received the reminder and fallback messages).

2. When a request comes in to `POST /` from Power Automate / Flow, this now counts it as only 0.5 of a button press. This will hopefully band-aid over the problem where it always **sends us two copies of each request**. This required a change in the DB to allow decimal values in the `num_presses` column, but I think it works. Unfortunately, at this time there are 38 rows in `sessions` that have an odd value in the `num_presses` column that are not the Princess Rooms (10 of them have `num_presses = 1`). This is out of the 1794 non-Princess Rooms sessions in that table.

3. **No longer require the `button-battery-level` header** in requests to `POST /flic_button_press`. I've seen cases where the Flic Internet Request doesn't send the button-battery-level header but everything else looks OK. I don't want to miss these presses, since there is no way to inform the resident that the button press did not result in a message. Therefore, this now allows these requests and just not updating the DB's button_battery_level value.

4. Now that we are counting each `POST /` as half a press. We can trust that the `num_presses` value is correct and should sent an **urgent text message when `numPresses == 2`**.

There are some other little changes to move things around, improve transaction handling, add/update tests, removing nearby semi-colons, etc.

This is available on `chatbot-dev` for testing.

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -29,7 +29,7 @@ const unrespondedSessionAlertTimeoutMillis = process.env.NODE_ENV === 'test' ? 2
 const accountSid = helpers.getEnvVar('TWILIO_SID');
 const authToken = helpers.getEnvVar('TWILIO_TOKEN');
 
-const client = require('twilio')(accountSid, authToken);
+const twilioClient = require('twilio')(accountSid, authToken);
 
 const heartbeatDashboardTemplate = fs.readFileSync(`${__dirname}/heartbeatDashboard.mst`, 'utf-8')
 const chatbotDashboardTemplate = fs.readFileSync(`${__dirname}/chatbotDashboard.mst`, 'utf-8')
@@ -72,7 +72,7 @@ async function handleValidRequest(button, numPresses) {
     }
 
     if(needToSendButtonPressMessageForSession(session)) {
-        sendButtonPressMessageForSession(session)
+        sendButtonPressMessageForSession(session, client)
     }
 
     await db.commitTransaction(client)
@@ -111,9 +111,8 @@ async function needToSendButtonPressMessageForSession(session) {
     return (session.numPresses === 1 || session.numPresses === 3 || session.numPresses % 5 === 0)
 }
 
-async function sendButtonPressMessageForSession(session) {
-
-    let installation = await db.getInstallationWithInstallationId(session.installationId)
+async function sendButtonPressMessageForSession(session, client) {
+    let installation = await db.getInstallationWithInstallationId(session.installationId, client)
 
     if (session.numPresses === 1) {
         await sendTwilioMessage(installation.responderPhoneNumber, session.phoneNumber, 'There has been a request for help from Unit ' + session.unit.toString() + ' . Please respond "Ok" when you have followed up on the call.')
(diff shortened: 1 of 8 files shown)


trisapeace:
Task linked: [CU-f0w9qm Improve logic of when to send initial and urgent text messages](https://app.clickup.com/t/f0w9qm)

server/server.js
```diff
@@ -294,25 +297,26 @@ app.get('/logout', (req, res) => {
...
-            }
-            else {
-                let numPresses = 1
+            } else {
+                await handleValidRequest(button, 1, batteryLevel)
 
-                if(req.query.presses == 2) {
-                    numPresses = 2;
+                if (req.query.presses == 2) {
+                    await handleValidRequest(button, 1)
```
trisapeace, line 317: @schwarrrtz @mariocimet 

This change is the only really important one. I think that we should get this deployed sooner rather than later because without it, we end up the situation where if the resident's first press is considered "numPresses = 2", then we don't send an initial message. It's only an issues at the Princess Rooms.

schwarrrtz:
re. "This means that when we received a request with `Type=double click`, we were treating it the same as a single click. This means that we might have missed the initial message for some urgent requests (they would still have received the reminder and fallback messages)."

good catch - but wouldn't they just have been downgraded to non-urgent presses? not lost entirely

server/server.js
```diff
@@ -339,19 +343,16 @@ app.post('/', jsonBodyParser, async (req, res) => {
...
                 res.status(400).send();
-            }
-            else {
-                let numPresses = 1
+            } else {
+                await handleValidRequest(button, 0.5)
 
-                if(req.body.Type == 'double_click') {
-                    numPresses = 2;
+                if (req.body.Type == 'double click') {
```
darvell, line 349: If things can sometimes be a bit iffy, why not req.body.Type.startsWith('double') so it's caught no matter what?

trisapeace:
@schwarrrtz You're right. I think I was getting it conflated with the problem of when `presses=2` comes in as the first request to `/flic_button_press`. Changed description to say the real consequences ("This means that we would not get an urgent message when we should have.")

server/server.js
trisapeace: Thanks @darvell , good idea. 

Done!

mariocimet (approved):

server/server.js
```diff
@@ -338,19 +337,16 @@ app.post('/', jsonBodyParser, async (req, res) => {
             if(button === null) {
                 log(`Bad request: UUID is not registered. UUID is ${req.body.UUID}`);
                 res.status(400).send();
-            }
-            else {
-                let numPresses = 1
+            } else {
```
trisapeace, line 346: @trisapeace Change this back.

schwarrrtz (approved):
After reviewing this myself & also in a meeting with Theresa and Mario I think we should approve this. I'd still like to test it myself on `chatbot-dev` before we deploy it. We also noted during the meeting that once we merge & deploy this we should get the old installations onto the Flic Internet Request system relatively soon.

server/server.js
trisapeace: Done!

schwarrrtz:
OK i've just finished up testing this on `chatbot-dev` and it seems to work great. Nice work @trisapeace.

PR #63, opened 2020-11-03
Use brave-alert-lib (CU-bar0fm)

trisapeace:
- Modified the code to use this version of `brave-alert-lib`: https://github.com/bravetechnologycoop/brave-alert-lib/pull/3
- Uses the `helpers` functions in `brave-alert-lib` wherever possible
- Moved the reminder and fallback timeout values into the `.env` file so that it can take advantage of the `helpers.getEnvVar()` method

- Also took this opportunity to remove some unused dependencies in `package.json` and to move some test-related dependencies into `devDependencies`


trisapeace:
Task linked: [CU-bar0fm Replace Chatbot Twilio code with BraveAlert](https://app.clickup.com/t/bar0fm)

server/package.json
```diff
@@ -5,10 +5,7 @@
   "main": "server.js",
   "dependencies": {
     "body-parser": "^1.18.3",
-    "chai": "^4.2.0",
-    "chai-http": "^4.2.0",
-    "chalk": "^2.4.2",
-    "connect-nedb-session": "0.0.3",
+    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#td-odetect-fixes",
```
trisapeace, line 8: This will need to be updating to the correct version after this PR is merged: bravetechnologycoop/brave-alert-lib#3

server/server.js
```diff
@@ -5,178 +5,195 @@ let https = require('https')
...
-    try {
-        await client.messages.create({from: fromPhoneNumber, body: message, to: toPhoneNumber})
-            .then(message => log(message.sid))
-    }
-    catch(err) {
-        log(err)
-    }
-}
+// Add BraveAlerter's routes
+app.use(braveAlerter.getRouter())
```
trisapeace, line 145: @schwarrrtz @mariocimet 
I remember you mentioning that DNS changes can sometimes take awhile. Do I need to do anything in the code to work around this change from `/message` to `/alert/sms`? Maybe have the one redirect to the other until the DNS change kicks in?

On the other hand, it looks like we can use this API call to change the SMS route for a given Twilio number: https://support.twilio.com/hc/en-us/articles/223136047-Configure-a-Twilio-Phone-Number-to-Receive-and-Respond-to-Messages#h_58646936-ad13-4940-aeb7-252b1d5aaa4e

I'm going to see if there's a way to get a list of all the Twilio numbers for an account.

trisapeace: OK, I managed to do this programmatically. I added the script to `utils.js` and tested it against the Buttons Dev Twilio account. What do you think? Can we avoid the DNS change if we do this during deployment?

schwarrrtz: Sounds like that script will be super useful during deployment!

The DNS changes we were talking about are at a higher level. It'll be something to consider when we change from `chatbot.brave.coop` to `buttons.brave.coop` or something like that. Other than updating the Twilio webhooks, we don't need to worry about changing the routes, as long as the FQDN stays the same

schwarrrtz: If we want to wait on updating the Twilio webhooks (eg. so that deployment is simpler) we could add a redirect from `/message` to `/alert/sms` like you're suggesting. But that's not related to DNS stuff.

trisapeace:
OK, this branch has now been rebased off `master` and deployed to `chatbot-dev`. Enjoy!

sampathsatti (approved):
MERGE ITTTT!

(Looks great, really love the new implementation - The code is self-explanatory as to its function :) )

server/BraveAlerterConfigurator.js
```diff
@@ -0,0 +1,126 @@
...
+            `There has been a request for help from Unit ${session.unit} . Please respond "Ok" when you have followed up on the call.`,
+            installation.responderPhoneNumber,
+            incidentCategoryKeys,
+            installation.incidentCategories,
+        )
+    
+        return alertSession
+    }
+
+    async alertSessionChangedCallback(alertSession) {
```
schwarrrtz, line 55: I'm confused about this and the associated addition of the db methods.

Is there a reason we can't use a single call to `db.saveSession` here? This would cut down on the number of db operations and require less new code.

Maybe I'm missing something though.

schwarrrtz (changes requested):

server/server.js
```diff
@@ -5,58 +5,38 @@ let https = require('https')
...
-    }
-}
+// Configure BraveAlerter
+const braveAlerter = (new BraveAlerterConfigurator()).createBraveAlerter()
 
-/**
-Check if a request is valid based on the presence of required body properties
-**/
-function isValidRequest(req, properties) {
+// Add BraveAlerter's routes
```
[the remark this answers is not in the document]
trisapeace: I totally agree. Done.

server/BraveAlerterConfigurator.js
trisapeace: Good question.

It's because of how the library is using it. Right now it's only setting the values of `alertSession` that it wants to change. For example, when it gets back any details/notes, then `alertSession.alertState` would be set to `COMPLETED` and `alertSession.details` would be set to, say, `'All good now'`. But it would leave the `alertSession.incidentCategory` as `undefined`. So if I were to use `db.saveSession` then I would overwrite `incidentCategory` in the DB with `null`, which is not what we want.

I agree that this is confusing. A couple of ways to change it so that it might be clearer would be:
* to have this function take in separate parameters for the values that can change instead of an incomplete `alertSession` object and otherwise leave the functionality the same with the separate DB calls.
* to have the library send back the whole session and then use `db.saveSession`; however, this means the library would need to learn how to do transactions using these callback functions (so there isn't a race condition between retrieving the values in the DB vs setting them to the new values).

What do you think?

schwarrrtz (approved):

server/BraveAlerterConfigurator.js
trisapeace: Change this to a transaction that 
1. Reads the session
2. Writes the whole session with the changes from `alertSession`

schwarrrtz:
oops, accidentally approved the wrong PR.

server/BraveAlerterConfigurator.js
trisapeace: Done.

server/package.json

server/BraveAlerterConfigurator.js
schwarrrtz: @trisapeace thanks.

schwarrrtz:
I think you mentioned that you're still testing the latest changes (?) but after a quick scan through the code it looks good.

schwarrrtz:
I think the new `alertSessionChangedCallback` is significantly easier to read and reason about, in addition to the other benefits of sticking to `db.saveSession`.

trisapeace:
I agree. I like this way better. Yes, I intend to put it on chatbot-dev and run it through its paces one more time before merging. But I'm hoping this is it :crossed_fingers:

trisapeace:
@schwarrrtz I'm finished my testing on `chatbot-dev`. I just now added a couple of null checks that were missing (https://github.com/bravetechnologycoop/BraveButtons/pull/63/commits/bbdb3bd963c4500866300186b7a844568181e84c) and I'm happy with it. I think it's good to be merged whenever you and @mariocimet approve it :smiley:

mariocimet (approved):

schwarrrtz (approved):
Looks great! Thanks for making the changes I requested.

PR #68, opened 2020-11-16
CU-fgyh0g adding battery level to sessions table and to dashboard 

mariocimet:
Tested on chatbot-dev, I've saved an extract of the registry column.

### DIFF
--- a/server/chatbotDashboard.mst
+++ b/server/chatbotDashboard.mst
@@ -197,6 +197,7 @@
                                 <th scope="col">Notes</th>
                                 <th scope="col">Started At</th>
                                 <th scope="col">Updated At</th>
+                                <th scope="col">Battery Level</th>
                             </tr>
                         </thead>
                         <tbody>
@@ -209,6 +210,7 @@
                                     <td>{{notes}}</td>
                                     <td>{{createdAt}}</td>
                                     <td>{{updatedAt}}</td>
+                                    <td>{{buttonBatteryLevel}}</td>
                                 </tr>
                             {{/recentSessions}}
                         </tbody>
--- a/server/db/010-addbatteryleveltosessions.sql
+++ b/server/db/010-addbatteryleveltosessions.sql
@@ -1,0 +1,21 @@
+DO $migration$
+    DECLARE migrationId INT;
+    DECLARE lastSuccessfulMigrationId INT;
+BEGIN
+    -- The migration ID of this file
+    migrationId := 10;
+
+    -- Get the migration ID of the last file to be successfully run
+    SELECT MAX(id) INTO lastSuccessfulMigrationId
+    FROM migrations;
+
+    -- Only execute this script if its migration ID is next after the last successful migration ID
+    IF migrationId - lastSuccessfulMigrationId = 1 THEN
+        ALTER TABLE sessions ADD COLUMN button_battery_level integer;
+        ALTER TABLE registry DROP COLUMN button_battery_level;
+
+        -- Update the migration ID of the last file to be successfully run to the migration ID of this file
+        INSERT INTO migrations (id)
+        VALUES (migrationId);
+    END IF;
+END $migration$;
--- a/server/db/db.js
+++ b/server/db/db.js
@@ -22,7 +22,7 @@ types.setTypeParser(types.builtins.NUMERIC, value => {
 })
 
 function createSessionFromRow(r) {
-    return new SessionState(r.id, r.installation_id, r.button_id, r.unit, r.phone_number, r.state, r.num_presses, r.created_at, r.updated_at, r.incident_type, r.notes, r.fallback_alert_twilio_status)
+    return new SessionState(r.id, r.installation_id, r.button_id, r.unit, r.phone_number, r.state, r.num_presses, r.created_at, r.updated_at, r.incident_type, r.notes, r.fallback_alert_twilio_status, r.button_battery_level)
 }
 
 function createInstallationFromRow(r) {
@@ -159,14 +159,14 @@ module.exports.getAllSessions = async function(client) {
     return rows.map(createSessionFromRow)
 }
 
-module.exports.createSession = async function(installationId, buttonId, unit, phoneNumber, numPresses, client) {
+module.exports.createSession = async function(installationId, buttonId, unit, phoneNumber, numPresses, buttonBatteryLevel, client) {
     let transactionMode = (typeof client !== 'undefined')
     if(!transactionMode) {
         client = await pool.connect()
     }
     
-    const values = [installationId, buttonId, unit, phoneNumber, ALERT_STATE.STARTED, numPresses]
-    const { rows } = await client.query('INSERT INTO sessions (installation_id, button_id, unit, phone_number, state, num_presses) VALUES ($1, $2, $3, $4, $5, $6) RETURNING *', values)
+    const values = [installationId, buttonId, unit, phoneNumber, ALERT_STATE.STARTED, numPresses, buttonBatteryLevel]
+    const { rows } = await client.query('INSERT INTO sessions (installation_id, button_id, unit, phone_number, state, num_presses, button_battery_level) VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING *', values)
    
     if(!transactionMode) {
         client.release()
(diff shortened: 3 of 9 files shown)


trisapeace:
Task linked: [CU-fgyh0g Add battery level alerting/display](https://app.clickup.com/t/fgyh0g)

server/db/009-addbatteryleveltosessions.sql
```diff
@@ -0,0 +1,20 @@
+DO $migration$
```
trisapeace, line 1: You're going to end up in conflict with @risatronic 's DB migration. You'll need to end up changing this to `migrationId := 10` and the file name change accordingly.

server/server.js
```diff
@@ -63,17 +63,19 @@ async function handleValidRequest(button, numPresses, batteryLevel) {
...
         session.incrementButtonPresses(numPresses)
+
         await db.saveSession(session, client)
     }
 
     if ((batteryLevel !== undefined) &&
             (batteryLevel >= 0) &&
             (batteryLevel <= 100)) {
         await db.saveButtonBatteryLevel(button.button_serial_number, batteryLevel, client)
+        await db.updateSessionBatteryLevel(session.id, batteryLevel, client)
```
trisapeace, line 78: Do you need this? Isn't it included in the `db.saveSession` call on line 71?

server/server.js
```diff
@@ -63,17 +63,19 @@ async function handleValidRequest(button, numPresses, batteryLevel) {
     let session = await db.getUnrespondedSessionWithButtonId(button.button_id, client)
         
     if (session === null) {
-        session = await db.createSession(button.installation_id, button.button_id, button.unit, button.phone_number, numPresses, client)
+        session = await db.createSession(button.installation_id, button.button_id, button.unit, button.phone_number, numPresses, batteryLevel, client)
```
trisapeace, line 66: Maybe this needs to be moved below the 
```
    if ((batteryLevel !== undefined) &&
            (batteryLevel >= 0) &&
            (batteryLevel <= 100)) {
```
condition so that you can find out whether the value of `batteryLevel` is even valid to put into the DB. If it's not, then you can save `null` in that column.

server/test/serverTest.js
```diff
@@ -350,20 +351,31 @@ describe('Chatbot server', () => {
...
 
             response = await chai.request(server).post('/flic_button_press').set('button-serial-number', unit1SerialNumber).set('button-battery-level', '0').send()
+            sessions = await db.getAllSessionsWithButtonId(unit1UUID)
+            session = sessions[0]
             button = await db.getButtonWithSerialNumber(unit1SerialNumber)
             expect(button.button_battery_level).to.equal(0)
+            expect(session.buttonBatteryLevel).to.equal(0)
         })
 
         it('should not change battery level column if subsequent requests do not provide button-battery-level', async () => {
```
trisapeace, line 381: Thanks for adding to the tests, could you add the session part to this one and the three other battery-related tests in this file?

trisapeace (commented):
Looks good. Makes sense to me as an approach.

I agree that it now makes the column in registry redundant. I'd be cool with you taking it out. After all, do we want two sources of truth? 

The one advantage I see that we get by leaving the column in `registry`is that we have a single place to look for all the most-recent battery levels. Although I imagine that we could grab that from `sessions` with some more interesting SQL.

server/server.js
mariocimet: oh! I hadn't accounted for the possibility that we could be saving -1 or some other non-null and also non-valid battery value, that's a good point

trisapeace:
Thinking more about the registry column. It would be unfortunate to lose the values that we have in there right now. Before removing that column, would you be able to move that value to their newest corresponding session row (or all their current session rows would also probably be accurate enough at this point).

server/db/db.js
```diff
@@ -521,6 +521,28 @@ module.exports.saveButtonBatteryLevel = async function(serialNumber, batteryLeve
     }
 }
 
+module.exports.updateSessionBatteryLevel = async function(sessionid, batteryLevel, client){
```
schwarrrtz, line 524: I have the same feedback here as I did for @trisapeace's BraveAlert integration PR - I don't think we need an extra db function for this, as it is already covered by `db.saveSession()`.

schwarrrtz (changes requested):
I agree with all of @trisapeace's feedback, and I've added an additional comment.

schwarrrtz:
@trisapeace regarding preserving the existing battery level data from the `registry` - I think it would be ok to back it up or export it, and compare with the new values as they come in to catch any inconsistencies. It might get weird trying to associate the registry battery level data with old sessions, and I don't know if it's worth the extra work. Jusy my two cents.

trisapeace:
@schwarrrtz I'd be OK with just exporting / backing it up.

server/server.js
```diff
@@ -41,19 +41,28 @@ async function handleValidRequest(button, numPresses, batteryLevel) {
...
-    else {
-        session.incrementButtonPresses(numPresses)
-        await db.saveSession(session, client)
-    }
 
     if ((batteryLevel !== undefined) &&
             (batteryLevel >= 0) &&
             (batteryLevel <= 100)) {
-        await db.saveButtonBatteryLevel(button.button_serial_number, batteryLevel, client)
+        if (session === null) {
```
trisapeace, line 48: This looks like it'll work, but I feel like the logic could be cleaned up somewhat to make it clearer and reduce repetition. Maybe by doing the logic to know whether we have a good batteryLevel or not first and then you can use that to do the create or update based on whether sessions is null.

trisapeace (approved):
Looks good to me.

I'd like to try it out on chatbot-dev after Shawna and I are finished with chatbot-dev's currently configuration.

This data will be so useful. My little button at home is still claiming 100 battery and I don't believe it.

server/server.js
mariocimet: do you mean assigning the battery level by itself and then doing the null check?

trisapeace: Yes, something like that. But no big deal if you don't want to fiddle with this anymore.

server/chatbotDashboard.mst
```diff
@@ -201,6 +203,7 @@
                                 <td>{{notes}}</td>
                                 <td>{{createdAt}}</td>
                                 <td>{{updatedAt}}</td>
+                                <td>{{buttonBatteryLevel}}</td>
```
trisapeace, line 206: This variable should match the name in the `viewParams` object. So either this should be `{{batteryLevel}}` or it should be `buttonBatteryLevel` in `viewParams`.

trisapeace (changes requested):

server/server.js
```diff
@@ -41,19 +41,28 @@ async function handleValidRequest(button, numPresses, batteryLevel) {
...
+            session = await db.createSession(button.installation_id, button.button_id, button.unit, button.phone_number, numPresses, batteryLevel, client)
+        }
+        else {
+            session.incrementButtonPresses(numPresses)
+            session.updateBatteryLevel(batteryLevel)
+            await db.saveSession(session, client)
```
trisapeace, line 54: It looks like we're now updating the battery level in *only* the session. In this case, I think we should remove the column from the `registry` table. Otherwise it'll get out of date and it will be confusing.

Either that, or both tables should be kept up-to-date.

mariocimet: added a drop column on registry to the migration script

server/server.js
mariocimet: I went to do this but then realized that I would need to change more than just this bit to ensure that we don't update the sessions table with a null value and overwrite the last useful battery value (if for some reason one press sends the battery level and the next one doesn't). So I'm leaving this as is for now.

mariocimet:
@trisapeace thank you for catching that bug! I think everything should be good now - the latest is on chatbot-dev

server/db/010-addbatteryleveltosessions.sql
```diff
@@ -0,0 +1,21 @@
...
+    migrationId := 10;
+
+    -- Get the migration ID of the last file to be successfully run
+    SELECT MAX(id) INTO lastSuccessfulMigrationId
+    FROM migrations;
+
+    -- Only execute this script if its migration ID is next after the last successful migration ID
+    IF migrationId - lastSuccessfulMigrationId = 1 THEN
+        ALTER TABLE sessions ADD COLUMN button_battery_level integer;
+        ALTER TABLE registry DROP COLUMN button_battery_level;
```
trisapeace, line 15: Thanks! Could you please make sure that we have an extract of that battery data that's already there so that we have it for buttons which aren't pressed after this change?

I think this can be done anytime now. I don't think there's much benefit in waiting until the deployment since we're only going to get one value from this anyways. What do you think?

server/server.js
```diff
@@ -41,19 +41,28 @@ async function handleValidRequest(button, numPresses, batteryLevel) {
...
-    }
-    else {
-        session.incrementButtonPresses(numPresses)
-        await db.saveSession(session, client)
-    }
 
     if ((batteryLevel !== undefined) &&
             (batteryLevel >= 0) &&
             (batteryLevel <= 100)) {
-        await db.saveButtonBatteryLevel(button.button_serial_number, batteryLevel, client)
```
trisapeace, line 56: You can probably delete `saveButtonBatteryLevel` from `db.js` now.

server/server.js
trisapeace: OK. That's fine. Thanks for looking into it.

server/server.js
trisapeace: Could you please make sure that this runs with the new change on `chatbot-dev`? I think just removing 10 from the `migrations` table and re-running `setup_server.sh` should do it.

trisapeace (approved):

server/db/010-addbatteryleveltosessions.sql
mariocimet: I put the latest on chatbot-dev and tested it with Postman, feel free to check out. I also copied the registry table to a csv on my machine

server/server.js
mariocimet: yay red ink red ink red ink red ink *chanting*

trisapeace (approved):
Looks great.

schwarrrtz:
@mariocimet you mentioned that you saved a copy of the `registry` table - was that from the dev server or the prod server?

schwarrrtz (approved):
I've tested this using Postman on `chatbot-dev`, and I ran the tests locally, which all pass except for the flaky fallback text test. Looks good.

risatronic (approved):

PR #71, opened 2020-12-01
setup_pi.sh and associated template files support network over PoE (CU-gcuhbk)

schwarrrtz:


trisapeace:
Task linked: [CU-gcuhbk add support for Brave Hub network connectivity over PoE](https://app.clickup.com/t/gcuhbk)

pi/interfaces.txt
```diff
@@ -1,4 +1,6 @@
 # refer to https://wiki.debian.org/BridgeNetworkConnectionsProxyArp
+# FLIC_HUB_INTERFACE is the interface connecting the Flic Hub to the RPi (either eth0 or eth1)
```
trisapeace, line 2: Would you be able to specify what `eth0` vs `eth1` means here?

And what `wlan0` vs `eth0` means in the next line? (I assume that you use `wlan0` for if you want wireless Internet and `eth0` if you want to use ethernet Internet, but it would be nice to have it spelled out.)

Or maybe these options and meanings would be better as comments in the `template.pi_config.ini` since that's actually where someone would need to decide which to use.

pi/interfaces.txt
```diff
@@ -9,13 +11,13 @@ iface lo inet loopback
...
 
-auto wlan0
-iface wlan0 inet dhcp
-wpa-conf /etc/wpa_supplicant/wpa_supplicant.conf
-  post-up /sbin/ip link set wlan0 promisc on
-  post-up /usr/sbin/parprouted FLIC_HUB_INTERFACE wlan0
+auto NETWORK_INTERFACE
+iface NETWORK_INTERFACE inet dhcp
+WPA_CONF
+  post-up /sbin/ip link set NETWORK_INTERFACE promisc on
```
trisapeace, line 17: Does this indentation represent anything functional based on `WPA_CONF`?

trisapeace (approved):
I'm really glad that you gave us a runthrough the other day about this template/replacement scheme. It took me a bit to figure out that that's what I was seeing here.

Everything looks fine to me.

mariocimet (approved):

pi/interfaces.txt
schwarrrtz: so `/etc/network/interfaces` has its own syntax and I'm not familiar enough with it to know the answer to this. what I do know is that all the example code that I've seen has this indentation scheme

pi/interfaces.txt
schwarrrtz: `eth0` is the builtin ethernet interface on the RPi. `eth1` is the USB-to-ethernet adapter, if present. `wlan0` is the wifi interface.

I've added some brief documentation to `pi/templates/interfaces.txt` explaining this.

pi/interfaces.txt
trisapeace: OK. I was just curious whether the indented lines should still appear when `WPA_CONF` is the empty string.

pi/interfaces.txt
trisapeace: Thanks!

pi/interfaces.txt
schwarrrtz: my understanding is that the `post-up` and `post-down` hooks are part of the `iface` stanza along with the `wpa-conf` (if present). welcome to weird linux networking syntax.

from the interfaces manpage (run `man interfaces` on a RPi):

```
The following "command" options are available for every family and method.  Each of these options can be given multiple times in a single stanza, in which case the commands are executed in the order in which they appear in the stanza.

pre-up ...
```

pi/setup_pi.sh
```diff
@@ -39,38 +41,67 @@ else
...
+    wpa_supplicant_config_file="${wpa_supplicant_config_file//FALLBACK_PSK/$fallback_psk}"
+    wpa_supplicant_config_file="${wpa_supplicant_config_file//SSID/$ssid}"
+    wpa_supplicant_config_file="${wpa_supplicant_config_file//PSK/$psk}"
+    echo "$wpa_supplicant_config_file" > /etc/wpa_supplicant/wpa_supplicant.conf
+  fi
 
-  interfaces_file=$(<$BASEDIR/interfaces.txt)
+  # set up the interfaces file (defines all network interfaces for the system)
+  # refer to https://wiki.debian.org/BridgeNetworkConnectionsProxyArp
+  interfaces_file=$(<$BASEDIR/templates/interfaces.txt)
```
trisapeace, line 75: I like this move to put the templates into a `/templates` folder. It makes their purpose clearer and the folder more organized.

pi/setup_pi.sh
```diff
@@ -39,38 +41,67 @@ else
...
-  echo 'DHCPHELPER_OPTS="-b wlan0"' > /etc/default/dhcp-helper
-  
+ 
+  # set up log files for the heartbeat script to use
   mkdir -p /var/log/brave
   touch /var/log/brave/heartbeat-out.log
   touch /var/log/brave/heartbeat-err.log
   chown pi:pi /var/log/brave /var/log/brave/heartbeat-out.log /var/log/brave/heartbeat-err.log
 
+  # set up a file to store the system_id (so it persists)
```
trisapeace, line 92: Thanks for adding all the comments!

trisapeace (approved):
Checked as well as I could with my little knowledge in this area.

PR #72, opened 2020-12-03
Add API key logging to `/flic_button_press` endpoint (CU-gwxnde)

trisapeace:
- To be used to verify that all the Buttons are correctly configured
  with the right API key before enforcing this authorization

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -221,10 +221,21 @@ app.post('/flic_button_press', Validator.header(['button-serial-number']).exists
         if(validationErrors.isEmpty()){
             const serialNumber = req.get('button-serial-number')
             const batteryLevel = req.get('button-battery-level')
+            const buttonName = req.get('button-name')
+            const apiKey = req.query.apikey
+
+            // Log the vaiditiy of the API key
+            // TODO (CU-gwxnde) Replace this with a 401 unauthorized if invalid
+            if (apiKey !== helpers.getEnvVar('FLIC_BUTTON_PRESS_API_KEY')) {
+                helpers.log(`INVALID api key from '${buttonName}' (${serialNumber})`)
+            }
+            else {
+                helpers.log(`VALID api key from '${buttonName}' (${serialNumber})`)
+            }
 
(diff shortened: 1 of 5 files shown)


trisapeace:
Task linked: [CU-gwxnde Chatbot sessions should not be able to be started unless by one of our Buttons](https://app.clickup.com/t/gwxnde)

server/server.js
```diff
@@ -221,10 +221,21 @@ app.post('/flic_button_press', Validator.header(['button-serial-number']).exists
         if(validationErrors.isEmpty()){
             const serialNumber = req.get('button-serial-number')
             const batteryLevel = req.get('button-battery-level')
+            const buttonName = req.get('button-name')
+            const apiKey = req.query.apikey
+
+            // Log the vaiditiy of the API key
+            // TODO (CU-gwxnde) Replace this with a 401 unauthorized if invalid
```
mariocimet, line 228: the client that would be receiving this is the Flichub itself right? So right now it can't do anything with a 401 if it did receive it? (Or I guess a spoofer would also be receiving this if they tried to spoof without an API key?)

mariocimet: and to be clear - right now we're not rejecting any requests with this? Just logging?

trisapeace: Yes, yes, yes, and yes :smile:

trisapeace: And yes. (I think I got all the question marks now, hehe).

I'm just logging for now because we can't actually reject requests until every button is set up with the key.

mariocimet (approved):

schwarrrtz:
Ok so from what I can tell we're just logging the API key for now? Then I assume we add some validation once all the installations have been upgraded to include the key? @trisapeace 

Other than updating the variable name to make it more explicit that we're generating the key ourselves, this all looks great.

trisapeace:
Yes, that's exactly right @schwarrrtz :-)

schwarrrtz (approved):

PR #73, opened 2020-12-15
Add dashboard landing page (CU-gurbtf)

risatronic:
Add landing page (/dashboard) which asks user to select an installation if any are loaded, otherwise informs them that none are


trisapeace:
Task linked: [CU-gurbtf make chatbot dashboard load even when there are no installations](https://app.clickup.com/t/gurbtf)

PR #74, opened 2020-12-15
Fixed Internet Request pairing instructions

mariocimet:
noticed this while doing manual testing yesterday - flic_button_request in the URL is a 404 Error

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,8 +245,7 @@ When performing a migration, make sure to connect directly to the target databas
 
 # How to pair a button with the 'Internet Request' Flic action
 
-1. Once you've run the pairing tool and button registering scripts, pair the button and select 'Internet Request' from the list of Flic actions. NOTE: this method will *not work from a button that has been paired to a smartphone*. *A Flic hub is required to make this work*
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -248,5 +248,5 @@ When performing a migration, make sure to connect directly to the target databas
 1. Once you've run the pairing tool and button registering scripts, pair the button and select 'Internet Request' from the list of Flic actions. NOTE: this method will *not work from a button that has been paired to a smartphone*. *A Flic hub is required to make this work*
 
 2. Select the POST request button, and input the url as follows:
```
trisapeace, line 250: Oh good! I'm glad that you found this @mariocimet ! I have not at all been keeping it up-to-date. Because right now:

* We aren't using the `?presses` query parameter anymore
* We are only configuring the "Click" action (because Double Click and Hold both fallback to the Click Action is a logical way)
* We need to add the `?apikey=` query parameter with the value from `FLIC_BUTTON_PRESS_API_KEY` in `.env`.

Is it OK if I add these changes as a commit to your branch/PR?

mariocimet: Yes please do!

trisapeace: Done. What do you think?

mariocimet: looks good to me!

trisapeace (approved):

schwarrrtz:
looks good.

schwarrrtz (approved):

PR #77, opened 2020-12-31
Add comments to the .env.example file

trisapeace:
- To make it clearer what values need to be in order for tests to pass


risatronic (approved):

schwarrrtz:
Looks good, I guess it does make sense for everyone to be using the same test timing. I still feel like the tests should be robust enough to handle that, and that maybe the fact that they aren't is why we're having some issues with flaky tests. Also the tests seem to work on my computer with different values for those env variables. But let's merge this for now.

schwarrrtz (approved):

PR #79, opened 2021-01-04
Added WiFi link quality monitoring to Brave Hub in heartbeat.py (CU-fmy630)

schwarrrtz:
- if `network_interface=wlan0` in `pi_config.ini`, the heartbeat script parses the output of `iwconfig` to determine the link quality, then logs it to `heartbeat.log`
- I tested this method by connecting a RPi to my home WiFi, starting the heartbeat script in a terminal window, then placing various metal objects between the RPi and my WiFi router. Without any objects the link quality was consistently reported as 1.0, with a thin sheet of tinfoil the link quality was around ~0.95, and with an aluminum Macbook Pro the link quality was ~0.9
- when adding tests for this feature, I moved all the sample darkstat HTML files used in existing tests to a new folder called `test_files` where I also placed some sample iwconfig output files.


trisapeace:
Task linked: [CU-fmy630 Add WiFi signal strength monitoring to Brave Hub](https://app.clickup.com/t/fmy630)

pi/heartbeat.py
```diff
@@ -102,6 +103,33 @@ def ping(host):
     logging.info('returned: {}'.format(returncode))
     return returncode == 0
 
+def parse_link_quality_from_iwconfig_output(iwconfig_output_text):
+    lines = iwconfig_output_text.split('\n')
+    for i in range(0, len(lines)):
+        # note: hardcoded wireless network interface name
+        if ('wlan0' in lines[i]) and (i+5 < len(lines)):
```
trisapeace, line 110: Instead of hardcoding it here, you could do it in the command itself and only have to deal with a single line (https://manpages.debian.org/buster/wireless-tools/iwconfig.8.en.html)

On my computer:
image

pi/heartbeat.py
```diff
@@ -102,6 +103,33 @@ def ping(host):
...
+    return -1.0
+
+def log_wifi_link_quality():
+    try:
+        iwconfig_output_text = subprocess.check_output('iwconfig').decode('utf-8')
+        link_quality = parse_link_quality_from_iwconfig_output(iwconfig_output_text)
+        if link_quality < 0:
+            logging.warning("error parsing iwconfig output")
+        else:
+            logging.info('wlan0 link quality is %f', link_quality)
```
trisapeace, line 129: Why not just output the whole `Link Quality ... Signal level.... ` line?

pi/test_heartbeat.py
```diff
@@ -42,18 +43,38 @@ def test_log_last_seen_when_darkstat_html_contains_flic_mac_address(self):
...
     def test_log_ip_when_darkstat_html_contains_flic_mac_address(self):
-        file_path = os.path.dirname(__file__) + '/sample_darkstat_html/log_ip.html'
+        file_path = os.path.dirname(__file__) + '/test_files/sample_darkstat_html/log_ip.html'
         with open(file_path, 'r') as html_file, unittest.mock.patch('heartbeat.logging') as mock_logging:
             html = html_file.read()
             heartbeat.parse_flic_ip_from_darkstat_html(html, '00:00:00:00:00:02')
 
             mock_logging.info.assert_called_once_with('darkstat html contained flic IPv4 address:  <td><a href="./192.168.8.114/">192.168.8.114</a></td>  <td>flic</td>  <td><tt>00:00:00:00:00:02</tt></td>  <td class="num">1 hr, 3 mins, 26 secs</td></tr>')
 
+class Test__parse_link_quality_from_iwconfig_output(object):
```
trisapeace, line 58: Thanks for writing tests :heart: 

Might be worth checking to see whether the string is logged, not just that the value is calculated correctly though.

trisapeace (commented):
Cool! Thanks so much for doing this. I hope that it gives us some good information.

pi/heartbeat.py
schwarrrtz: Ah yeah good idea. I'll do this.

pi/heartbeat.py
schwarrrtz: I was thinking of wanting to graph this in the future, so it made sense to normalize the Link Quality fraction that `iwconfig` reports. I don't think the denominator is always the same, although it was always 70 for me, and sounds like it's the same for you on your machine.

The signal level is somewhat useful but for me it wasn't logging the noise level, and what we really want to know is the SNR, so the signal level alone is less useful. The `iwconfig` docs seem to indicate that the noise level will be logged, but I don't see it. That was my thinking anyway. And from my manual testing it seemed like the Link Quality metric was changing in response to real world things, so it seemed like a good place to start.

pi/test_heartbeat.py
schwarrrtz: The issue with this is that if `iwconfig` isn't present on the test machine, the test will fail. I guess the test could check for the presence of `iwconfig` and then not run? But I feel like this will often be the case - ie. I doubt the Travis virtual boxes have a wlan0 interface, `iwconfig` doesn't work on my WSL dev environment, etc.

schwarrrtz: I guess I could also mock the `subprocess.check_output` call, but at that point the additional tests would basically just be 'does the logging library work'

schwarrrtz: What do you think about all this @trisapeace?

pi/heartbeat.py
trisapeace: Looking at this Stackoverflow answer (https://superuser.com/questions/866005/wireless-connection-link-quality-what-does-31-70-indicate), it looks like the denominator will always be 70. Whether to trust or not, is always the question.

For graphing, makes sense. I just figured since you were logging a string that you'd need to parse out the number from afterwards anyways, it seemed like less work to just use the string they already give you. Up to you!

pi/test_heartbeat.py
trisapeace: What do you think of the `mock_logging.info.assert_called_once_with(...)` construct used in the test directly above you new ones?

schwarrrtz: ok I've updated the implementation so that this test structure is possible, and changed the tests as well.

pi/heartbeat.py
schwarrrtz: done.

pi/heartbeat.py
schwarrrtz: Hm, interesting post. The `iw` tool they mention yells not to 'screenscrape' it as the output format isn't stable apparently.

Fair point on the logging, I remember struggling to get the log files into the formats that I wanted them in for working with the Flic last seen values and wishing they were formatted in a more friendly way in the logs themselves. Hence my decision to implement this feature the way that I did.

trisapeace (approved):
Looks good to me!

risatronic (approved):

mariocimet (approved):

PR #81, opened 2021-01-15
ESLint Preview

risatronic:
Putting in PR not intending to be merged


schwarrrtz:
I'm going to need a bit of time to look through all this, but my first impression is that I'd be happier if we could fine tune the rules a bit. I really don't like the way it splits function arguments across so many lines, for example.

Maybe we should have a meeting about this after all

risatronic:
> I'm going to need a bit of time to look through all this, but my first impression is that I'd be happier if we could fine tune the rules a bit. I really don't like the way it splits function arguments across so many lines, for example.
> 
> Maybe we should have a meeting about this after all

No huge rush I don't think, I mostly just wanted to get it to a place where people could see what exactly it's doing and go from there :) I'm up for a meeting though

risatronic:
It did make some files in both repos considerably longer which isn't my favourite either

PR #82, opened 2021-01-20
Handle all promise rejections in async functions (CU-brqahw)

trisapeace:
- Added try/catches to all the async functions. This should handle all our promise rejections.
   - At the moment, the `catch` blocks generally don't do anything interesting. They just log the error. This follows what @mariocimet did in the BraveSensor-Server codebase. This would probably be a good place to introduce Sentry into our code. Then we can be notified when these happen in production.
- Added rollbacks in the `catch` blocks of functions that use DB transactions. This might help remove the deadlock problems that we've been seeing when running the tests.

I tried writing some automated tests for this, but as soon as I'd try to catch the promise rejection, then it would not longer be unhandled and hence the error would go away. So I just tried a few things out manually to see if the warning was there or not when I expected it to be.

As of Jan 19, this is deployed on `chatbot-dev`. I've tried out all the endpoints and they still seem to work as I expected.

### DIFF
--- a/server/BraveAlerterConfigurator.js
+++ b/server/BraveAlerterConfigurator.js
@@ -13,50 +13,63 @@ class BraveAlerterConfigurator {
     }
 
     async getAlertSession(sessionId) {
-        const session = await db.getSessionWithSessionId(sessionId)
-        if (session === null) {
-            return null
-        }
+        let alertSession = null
+        try {
+            const session = await db.getSessionWithSessionId(sessionId)
+            if (session === null) {
+                return null
+            }
 
-        const installation = await db.getInstallationWithInstallationId(session.installationId)
-    
-        const incidentCategoryKeys = this.createIncidentCategoryKeys(installation.incidentCategories)
-    
-        let alertSession = new AlertSession(
-            session.id,
-            session.state,
-            session.incidentType,
-            session.notes,
-            `There has been a request for help from Unit ${session.unit} . Please respond "Ok" when you have followed up on the call.`,
-            installation.responderPhoneNumber,
-            incidentCategoryKeys,
-            installation.incidentCategories,
-        )
+            const installation = await db.getInstallationWithInstallationId(session.installationId)
+        
+            const incidentCategoryKeys = this.createIncidentCategoryKeys(installation.incidentCategories)
+        
+            alertSession = new AlertSession(
+                session.id,
+                session.state,
+                session.incidentType,
+                session.notes,
+                `There has been a request for help from Unit ${session.unit} . Please respond "Ok" when you have followed up on the call.`,
+                installation.responderPhoneNumber,
+                incidentCategoryKeys,
+                installation.incidentCategories,
+            )
+        }
+        catch (e) {
+            helpers.log(`getAlertSession: failed to get and create a new alert session: ${JSON.stringify(e)}`)
+        }
     
         return alertSession
     }
 
     async getAlertSessionByPhoneNumber(toPhoneNumber) {
-        const session = await db.getMostRecentIncompleteSessionWithPhoneNumber(toPhoneNumber)
-        if (session === null) {
-            return null
-        }
+        let alertSession = null
 
-        const installation = await db.getInstallationWithInstallationId(session.installationId)
-    
-        const incidentCategoryKeys = this.createIncidentCategoryKeys(installation.incidentCategories)
-    
-        let alertSession = new AlertSession(
-            session.id,
-            session.state,
-            session.incidentType,
-            session.notes,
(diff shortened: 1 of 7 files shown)


trisapeace:
Task linked: [CU-brqahw Handle all promise rejections in async functions](https://app.clickup.com/t/brqahw)

risatronic (approved):
I do feel confident in this area and these changes all look good to me! Adding rollbacks in particular seems like a good idea

mariocimet (approved):

server/db/db.js
```diff
@@ -34,488 +34,654 @@ function createHubFromRow(r) {
...
+        }
 
-    if(rows.length > 0) {        
-        return createSessionFromRow(rows[0])
+        const query = "SELECT * FROM sessions WHERE button_id = $1 AND state != $2 AND state != $3 AND state != $4"
+        const values = [buttonId, ALERT_STATE.WAITING_FOR_CATEGORY, ALERT_STATE.WAITING_FOR_DETAILS, ALERT_STATE.COMPLETED]
+        const { rows } = await client.query(query, values)
+    
+        if(!transactionMode) {
+            client.release()
```
schwarrrtz, line 88: if we're not in transaction mode, but an error gets thrown during the SELECT query above, the code path will skip this `client.release()` call. this might lead to our db client pool running out of clients to use.

schwarrrtz (changes requested):
Theresa and I just had a good discussion about this - there are some places where additional logging would be nice, and some places where the new `try/catch` structure might lead to the db pool running out of clients. We agreed that those issues should be resolved before this gets merged. But overall I think this will significantly improve our error handling & debugging capabilities. Thanks @trisapeace !

server/BraveAlerterConfigurator.js
```diff
@@ -13,80 +13,104 @@ class BraveAlerterConfigurator {
...
             }
 
-            await db.saveSession(session, client)
+            await db.commitTransaction(client)
+        } catch (e) {
+            try {
+                await db.rollbackTransaction(client)
+                helpers.log(`alertSessionChangedCallback: Rolled back transaction because of error: ${e}`)
+            } catch (error) {
+                // Do nothing
```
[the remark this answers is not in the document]
trisapeace: Done!

server/db/db.js
trisapeace: Done.

server/BraveAlerterConfigurator.js
```diff
@@ -13,80 +13,109 @@ class BraveAlerterConfigurator {
...
+        let client
 
-        const session = await db.getSessionWithSessionId(alertSession.sessionId, client)
-        
-        if (session) {
-            if (alertSession.alertState) {
-                session.state = alertSession.alertState
+        try {
+            client = await db.beginTransaction()
+            if (client === null) {
```
trisapeace, line 81: Also added this `null` handling here and in `server.js`.

server/db/db.js
```diff
@@ -34,169 +34,293 @@ function createHubFromRow(r) {
...
+        // this fixes a race condition when two button press messages are received in quick succession
+        // this means that only one transaction executes at a time, which is not good for performance
+        // we should revisit this when / if db performance becomes a concern
+        await client.query("LOCK TABLE sessions, registry, installations, migrations")
+    }
+    catch(e) {
+        helpers.log(`Error running the beginTransaction query: ${e}`);
+        if (client) {
+            try {
+                await this.rollbackTransaction(client)
```
trisapeace, line 52: I added this rollback attempt here too.

server/server.js
```diff
@@ -38,68 +38,90 @@ app.use(braveAlerter.getRouter())
 async function handleValidRequest(button, numPresses, batteryLevel) {
     helpers.log('UUID: ' + button.button_id.toString() + ' SerialNumber: ' + button.button_serial_number + ' Unit: ' + button.unit.toString() + ' Presses: ' + numPresses.toString() + ' BatteryLevel: ' + batteryLevel)
 
-    let client = await db.beginTransaction()
+    let client 
+    
+    try {
+        client = await db.beginTransaction()
+        if (client === null) {
+            helpers.log(`alertSessionChangedCallback: Error starting transaction`)
```
schwarrrtz, line 46: I think this log message needs to be updated?

schwarrrtz (changes requested):

server/server.js
trisapeace: Fixed. Thanks!

schwarrrtz (approved):
Thanks for making all the changes in response to my reviews @trisapeace ! Looks good now, I'm still not 100% following all the code paths and implications of making things nullable but I feel good about merging it at this point.

PR #84, opened 2021-02-01
Use `npm ci` instead of `npm install` for Travis and deployment scripts (CU-jcwffp)

trisapeace:
This means that they will always use the exact versions specified in the package-lock.json file, thus ensuring more consistent builds (see NPM docs: https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable)

Note that the Task only specifies to do this for `/server`. I also made the change for the `/pi` directory. I think we had talked about this before and came to the conclusion to only do `/server`, but now I can't remember why. Do you remember @schwarrrtz or @mariocimet ?

I also added the missing deployment date from the CHANGELOG that I forgot.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ matrix:
       # install dependencies
       install:
         - cd server
-        - npm install
+        - npm ci
 
       # install database
       addons:
@@ -78,7 +78,7 @@ matrix:
       # install dependencies
       install:
         - cd pairing-tool
-        - npm install
+        - npm ci
 
       # run linter
       script:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,10 @@ the code was deployed.
 ### Added
 - Implement linting on Buttons repo (CU-eprhhn).
 
-## [3.2.0]
+## Changed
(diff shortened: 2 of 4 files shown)


trisapeace:
Task linked: [CU-jcwffp Use `npm ci` instead of `npm install` in /server](https://app.clickup.com/t/jcwffp)

risatronic (approved):

risatronic:
> This means that they will always use the exact versions specified in the package-lock.json file, thus ensuring more consistent builds (see NPM docs: https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable)
> 
> Note that the Task only specifies to do this for `/server`. I also made the change for the `/pi` directory. I think we had talked about this before and came to the conclusion to only do `/server`, but now I can't remember why. Do you remember @schwarrrtz or @mariocimet ?
> 
> I also added the missing deployment date from the CHANGELOG that I forgot.

commenting because I'm curious as to why not /pi if that's true

mariocimet (approved):

schwarrrtz:
I don't see a reason why we shouldn't use `nom ci` for the pairing tool as well.

schwarrrtz (approved):

PR #85, opened 2021-02-12
Use helpers.logError instead of console.error (CU-jcuw85)

trisapeace:
- This will add the timestamp to error output in production

### DIFF
--- a/server/db/db.js
+++ b/server/db/db.js
@@ -15,8 +15,7 @@ const pool = new Pool({
 })
 
 pool.on('error', err => {
-  // eslint-disable-next-line no-console
-  console.error('unexpected database error:', err)
+  helpers.logError(`unexpected database error: ${JSON.stringify(err)}`)
 })
 
 types.setTypeParser(types.builtins.NUMERIC, value => parseFloat(value))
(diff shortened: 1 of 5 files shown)


trisapeace:
Task linked: [CU-jcuw85 add timestamps to server error logs](https://app.clickup.com/t/jcuw85)

mariocimet (approved):

schwarrrtz (approved):
looks good.

server/db/db.js
```diff
@@ -15,8 +15,7 @@ const pool = new Pool({
 })
 
 pool.on('error', err => {
-  // eslint-disable-next-line no-console
-  console.error('unexpected database error:', err)
+  helpers.logError('unexpected database error:', err)
```
trisapeace, line 18: Hmmm, actually now that I'm looking at this again, I think that the `logError` function only takes one parameter. So I probably need to make this all a single string. This is a reminder for me to check this before merging.

trisapeace: Yup, the function only takes in one argument, the string to output. So I've changed this now to do that.

trisapeace:
Waiting for https://github.com/bravetechnologycoop/brave-alert-lib/pull/8 .
Then I'll upgrade the version of `brave-alert-lib` and the timestamps will only appear once in the logs.

risatronic (approved):
Looks good to me!

mariocimet (approved):

PR #86, opened 2021-03-02
Add CSV Button (CU-c8htxp)

risatronic:
- Add CSV button to dashboard so people don't need to ask devs to get information for them from database
- Remove line that allows static file serving

~~It doesn't let you click anything else after the download, see [this message](https://bravely.slack.com/archives/C01KLE54K7Y/p1614662642004500)~~

~~Hopefully that's a quick fix and can be done in code review tomorrow~~

^ This is fixed


trisapeace:
Task linked: [CU-c8htxp Provide latest Master Press Data Sheet data on demand to Brave users](https://app.clickup.com/t/c8htxp)

server/server.js
```diff
@@ -252,6 +254,13 @@ app.get('/dashboard/:installationId?', async (req, res) => {
   }
 })
 
+app.get('/buttons-data', async (req, res) => {
+  const data = await db.getDataForExport()
+
+  const csv = csvParser.parse(data)
+  res.set('Content-Type', 'text/csv').attachment('buttons-data.csv').send(csv)
```
trisapeace, line 261: I suggest adding a timestamp to the filename.

server/server.js
```diff
@@ -252,6 +254,13 @@ app.get('/dashboard/:installationId?', async (req, res) => {
   }
 })
 
+app.get('/buttons-data', async (req, res) => {
```
trisapeace, line 257: I think this is worth running a few performance tests.

Right now we have about 2500 sessions in production. It would be great to try this out on 100, 1000, 5000, and 10000 sessions.

I'm mostly interested in these things:
* How long does it take?
* Does the time it takes grow linearly or exponentially?
* Are Button presses still being handled while this is running (assuming that it does run for long enough to try this out)

schwarrrtz:
please restrict access to this feature to logged in users :)

server/chatbotDashboard.mst
```diff
@@ -153,7 +153,7 @@
 
 <body>
     <nav class="navbar navbar-expand-md">
-        <a class="navbar-brand" id="navbar-brand">Brave Chatbot Dashboard</a>
+        <a class="navbar-brand" id="navbar-brand" href="/dashboard">Brave Chatbot Dashboard</a>
```
schwarrrtz, line 156: This is probably unnecessary, and might be confusing if people accidentally click the page title.

server/server.js
```diff
@@ -252,6 +254,13 @@ app.get('/dashboard/:installationId?', async (req, res) => {
...
+  const data = await db.getDataForExport()
+
+  const csv = csvParser.parse(data)
```
trisapeace, line 260: I'm not super worried about special characters because this uses a library, but it would probably still be worth putting a double-quote in the `notes` column and making sure that it exports it properly (I would expect it to put double-double-quotes (i.e instead of `Hello"World` it would do `Hello""World`).

risatronic: Tried this locally and it still works!

server/server.js
```diff
@@ -154,20 +156,20 @@ app.use((req, res, next) => {
```
trisapeace, line 156: Is the `app.use` middleware directly above this still needed?

trisapeace: Looking at it again, I think we do. Because it'll log people out when the cookie expires. So not a problem.

server/db/db.js
```diff
@@ -844,6 +844,36 @@ async function saveHubAlertStatus(hub, clientParam) {
...
+      client = await pool.connect()
+    }
+
+    const { rows } = await client.query(
+      `SELECT i.name AS "Installation Name", i.responder_phone_number as "Responder Phone", i.fall_back_phone_number AS "Fallback Phone", i.created_at AS "Date Installation Created", i.incident_categories AS "Incident Categories", i.is_active AS "Active?", s.button_id AS "Button ID", r.button_id AS "Button ID Registry - compare", s.unit AS "Unit", r.unit AS "Unit Registry - Compare", s.phone_number AS "Button Phone", r.phone_number AS "Button Phone Registry - compare", s.state AS "Session State", s.num_presses AS "Number of Presses", s.created_at AS 
+      "Session Start", s.updated_at AS "Last Session Activity", s.incident_type AS "Session Incident Type", s.notes as "Session Notes", s.fallback_alert_twilio_status AS "Fallback Alert Status (Twilio)", s.button_battery_level AS "Button Battery Level", r.created_at AS "Date Button Created", r.updated_at AS "Button Last Updated", r.button_serial_number AS "Button Serial Number" FROM installations i JOIN sessions s ON i.id = s.installation_id JOIN registry r ON i.id = r.installation_id`,
+    )
+
+    if (rows.length > 0) {
```
trisapeace, line 861: What if `rows.length` is zero? I think it should be OK to output a CSV with only the headers.

server/server.js
```diff
@@ -252,6 +254,17 @@ app.get('/dashboard/:installationId?', async (req, res) => {
   }
 })
 
+app.get('/buttons-data', sessionChecker, async (req, res) => {
+  const data = await db.getDataForExport()
+  const csv = csvParser.parse(data)
+
+  const millis = Date.now()
+  const timestamp = new Date(millis).toISOString().slice(0, -5).replace(/T|:/g, '_')
+  console.log(data)
```
trisapeace, line 263: I think this `console.log` should be removed.

server/db/db.js
risatronic: Yeah that makes sense actually, good call

risatronic: Fixed :)

server/server.js
risatronic: Gone!

risatronic:
I think all I have left to do is run the performance tests Theresa was mentioning now that I've gotten the particular query nailed down and checked the timestamp format with the folks who will be using it :)

server/db/db.js
trisapeace: Cool. Thanks! Did you check that this works when there are zero rows?

risatronic: I double-checked even though I've gotten a bunch of CSVs locally with no data and just headers leading up to now, but I guess they were all before I implemented the column renaming cause it broke 🤣 Added in a fields setting just now that avoids that :)

server/server.js
risatronic: This has been done, more info/conversation in this slack thread: https://bravely.slack.com/archives/C01KLE54K7Y/p1614972901043600

trisapeace (approved):
Looks nice to me.

mariocimet (approved):

PR #88, opened 2021-03-11
Update Button Press Handling (CU-j6yfct, CU-j6y98q)

risatronic:
This also includes [this task](https://app.clickup.com/t/j6y98q), I don't think they can both be linked in the title

- Start a new session if "current" session has been in "Waiting for reply" for >= 2 hours so that important presses are not missed if sessions are not completed
- Send additional alerts on presses that aren't a multiple of 5 if it has been more than 2 minutes since last press

**currently any subsequent text messages will cascade through to the last incomplete session(s), not sure that this is best to try to address here vs in brave-alert-lib so open to thoughts!


trisapeace:
Task linked: [CU-j6yfct Send initial text message if most recent session has been left in "Waiting for reply" for more than 2 hours](https://app.clickup.com/t/j6yfct)

risatronic:
This is up on dev if anyone feels like playing with it :)

server/server.js
```diff
@@ -88,9 +102,13 @@ async function handleValidRequest(button, numPresses, batteryLevel) {
...
 
     let currentSession = await db.getUnrespondedSessionWithButtonId(button.button_id, client)
+    const currentTime = await db.getCurrentTime(client)
 
     if (batteryLevel !== undefined && batteryLevel >= 0 && batteryLevel <= 100) {
-      if (currentSession === null) {
+      if (
+        currentSession === null ||
+        (currentTime - currentSession.updatedAt >= SESSION_RESET_TIMEOUT && currentSession.state === ALERT_STATE.WAITING_FOR_REPLY)
+      ) {
```
risatronic, line 111: I was looking at this again and questioning whether it needed to check the state but based on `getUnrespondedSessionWithButtonId`'s check I think it does? It is late though so I'm not sure my brain is working right on that 😅

```
const query = 'SELECT * FROM sessions WHERE button_id = $1 AND state != $2 AND state != $3 AND state != $4'
const values = [buttonId, ALERT_STATE.WAITING_FOR_CATEGORY, ALERT_STATE.WAITING_FOR_DETAILS, ALERT_STATE.COMPLETED]
```

trisapeace: You're right, `db.getUnrespondedSessionsWithButtonId` does check the state, so you don't need to check it here.

Also, whatever change you make to this `if`-statement also needs to be done on the one further down on line 126: `} else if (currentSession === null) {`. This happens in both these places because here is if we got a good battery level, and the other is if we didn't.

server/server.js
```diff
@@ -36,14 +38,22 @@ const braveAlerter = new BraveAlerterConfigurator().createBraveAlerter()
...
 
-function needToSendButtonPressMessageForSession(currentSession) {
-  return currentSession.numPresses === 1 || currentSession.numPresses === 2 || currentSession.numPresses % 5 === 0
+async function needToSendButtonPressMessageForSession(currentSession) {
+  const currentTime = await db.getCurrentTime()
+  return (
+    currentSession.numPresses === 1 ||
+    currentSession.numPresses === 2 ||
+    currentSession.numPresses % 5 === 0 ||
+    currentTime - currentSession.updatedAt >= REMINDER_MESSAGE_THRESHOLD
```
trisapeace, line 47: Can you please add a test that makes sure that this works?

server/server.js
```diff
@@ -59,7 +69,11 @@ async function sendButtonPressMessageForSession(currentSession, client) {
         fallbackFromPhoneNumber: helpers.getEnvVar('TWILIO_FALLBACK_FROM_NUMBER'),
       }
       braveAlerter.startAlertSession(alertInfo)
-    } else if (currentSession.numPresses % 5 === 0 || currentSession.numPresses === 2) {
+    } else if (
+      currentSession.numPresses % 5 === 0 ||
+      currentSession.numPresses === 2 ||
+      currentTime - currentSession.updatedAt >= REMINDER_MESSAGE_THRESHOLD
```
trisapeace, line 75: Can you please add test(s) to show this working?

server/server.js
```diff
@@ -88,9 +102,13 @@ async function handleValidRequest(button, numPresses, batteryLevel) {
     }
 
     let currentSession = await db.getUnrespondedSessionWithButtonId(button.button_id, client)
+    const currentTime = await db.getCurrentTime(client)
 
     if (batteryLevel !== undefined && batteryLevel >= 0 && batteryLevel <= 100) {
-      if (currentSession === null) {
+      if (
+        currentSession === null ||
+        (currentTime - currentSession.updatedAt >= SESSION_RESET_TIMEOUT && currentSession.state === ALERT_STATE.WAITING_FOR_REPLY)
```
trisapeace, line 110: Please write tests to show that this works for both the case where there is a batteryLevel and when there isn't.

trisapeace (commented):
Good start! This is going to be really useful. I'm getting tired of phoning buildings to ask them to respond to unresponded sessions 😬

risatronic:
> Good start! This is going to be really useful. I'm getting tired of phoning buildings to ask them to respond to unresponded sessions 😬

oh my gosh I didn't realise that was a thing 😲

server/server.js
risatronic: ooh right that makes sense!

server/server.js
```diff
@@ -10,14 +10,16 @@ const jsonBodyParser = bodyParser.json()
...
+const { ALERT_STATE, helpers } = require('brave-alert-lib')
 
 const BraveAlerterConfigurator = require('./BraveAlerterConfigurator.js')
 const db = require('./db/db.js')
 
 const FLIC_THRESHOLD_MILLIS = 210 * 1000
 const HEARTBEAT_THRESHOLD_MILLIS = 75 * 1000
 const PING_THRESHOLD_MILLIS = 320 * 1000
+const SESSION_RESET_TIMEOUT = 7200 * 1000
+const REMINDER_MESSAGE_THRESHOLD = 120 * 1000
```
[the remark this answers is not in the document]
risatronic: I wasn't sure if that was overkill or not, but it makes what it is more obvious to me too so I can definitely update that :)

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ was committed to the `master` branch. This is not necessarily the date that
 the code was deployed.
 
 ## [Unreleased]
+### Added
+- Send urgent messages if it has been more than 2 minutes since last button press, regardless of if it is a multiple of 5 presses (CU-j6y98q).
```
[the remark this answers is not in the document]
risatronic: Yes - do you think I should say since last session update maybe though since that is more accurate?

(Summary of why for other people: the reminder and fallback messages both also update the session so if someone presses for a 3rd or 4th time right after those updates those presses will be lost as far as having their own alerts goes, but there will still be 2 alerts 2 minutes apart in that 4 minute window and any presses that happen >= 2 minutes after the fallback will send an alert if it's been 2 minutes since the last one regardless of if they're a multiple of 5)

server/server.js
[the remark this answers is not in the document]
risatronic: Oh right, I also just realised looking at this more that I'm not sure the name of the 2nd one is descriptive for what it is, I'm thinking maybe `SUBSEQUENT_URGENT_MESSAGE_THRESHOLD` based on the task name in clickup ("Also send subsequent urgent request messages based on time since last message")? Cause it's not really a reminder in my mind if it's a new urgent message that says how many presses have happened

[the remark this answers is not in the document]
trisapeace: Very true. `SUBSEQUENT_URGENT_MESSAGE_THRESHOLD` works for me.

risatronic:
Trying to write up testing for this is driving me up the wall and I think part of why it's not working may be related to my local environment, maybe people can offer some insight during code review tomorrow 😓

risatronic:
This should be good to go now :)

server/db/db.js
```diff
@@ -870,6 +870,32 @@ async function getDataForExport(clientParam) {
...
+async function getCurrentTime(clientParam) {
+  let client = clientParam
+  const transactionMode = typeof client !== 'undefined'
+
+  try {
+    if (!transactionMode) {
+      client = await pool.connect()
+    }
+
+    const { rows } = await client.query('SELECT NOW()')
```
trisapeace, line 882: Did you mean to leave this as `NOW()`? Does it work on the server as well?

server/test/serverTest.js
```diff
@@ -584,6 +584,165 @@ describe('Chatbot server', () => {
...
+        .request(server)
+        .post('/flic_button_press')
+        .set('button-serial-number', unit1SerialNumber)
+        .set('button-battery-level', '100')
+        .send()
+
+      expect(imports.braveAlerter.sendSingleAlert).to.be.calledWith(
+        sinon.match.any,
+        sinon.match.any,
+        'This in an urgent request. The button has been pressed 3 times. Please respond "Ok" when you have followed up on the call.',
```
trisapeace, line 738: I like this. Good idea to check that the number increments the way we expect.

server/db/db.js
risatronic: Yes - this way I (or anyone else who doesn't have the same PSQL version as the server) don't need to change back and forth between `now` & `current_timestamp`, it's now across the board in both places now 😅

server/test/serverTest.js
```diff
@@ -584,6 +584,165 @@ describe('Chatbot server', () => {
...
+      // eslint-disable-next-line prettier/prettier
+      await chai
+        .request(server)
+        .post('/flic_button_press')
+        .set('button-serial-number', unit1SerialNumber)
+        .send()
+
+      expect(imports.braveAlerter.startAlertSession).to.have.been.calledTwice
+
+      db.getCurrentTime.restore()
```
trisapeace, line 617: So this is a tricky one. This line won't be executed if any of the `expect` tests fails, which will can cause some hard-to-debug test failures in other tests. Normally, I try to put my calls to `restore()` in the `afterEach` function so that I know that it will be executed no matter what. In this case, you probably don't want to change the existing `beforeEach` and `afterEach` functions. But what you could do is put your four new tests into a new `describe` that has its own `afterEach` to restore the stub. This new `describe` can be nested right here where you've got the `it` s, and that way it should still have the same `beforeEach` and `afterEach` as before, but then just also have your new one. Let me know if you have any questions!

risatronic: Done :)

trisapeace: Could you please push this change :smile:

risatronic: Oops yes 😓🤣

trisapeace: Thanks!

trisapeace:
Looks great. And I tried it on dev last night and it worked for me for both the 2 minutes for urgent messages and the 2 hours for starting a new session.

trisapeace (approved):

mariocimet (approved):

schwarrrtz (approved):
this looks good, and the tests all pass for me locally. nice work @risatronic :)

PR #89, opened 2021-03-27
Add ability to send multiple fallback messages (CU-pv8hd5, CU-p107rc, CU-307j4h)

trisapeace:
- Upgraded version of brave-alert-lib (https://github.com/bravetechnologycoop/brave-alert-lib/pull/10) takes an array of phone numbers for the fallbackPhoneNumbers parameter instead of just one.
- Changed `installations.fall_back_phone_number` column in the DB to be an text array, renamed it `fall_back_phone_numbers`, and migrated the current value in `fall_back_phone_number` into an array containing the old value.
- Updated tests accordingly
- Note that setup_server.sh 's script still only expects a single fallback number because at the point of running the migration file that needs that parameter, the column is still only expecting a text string containing a single phone number

- Tested this on chatbot dev:
   - The fallback messages are sent to multiple phones
   - Export CSV with an array of fallback phone numbers and the Twilio return values from multiple fallback message attempts
   - add_installation.sh script with an array of fallback phone numbers
   - Database migration with values in the DB

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Changelog
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
@@ -9,11 +10,14 @@ was committed to the `master` branch. This is not necessarily the date that
 the code was deployed.
 
 ## [Unreleased]
+
 ### Added
 - Send urgent message on button press if it has been >= 2 minutes since last session update, regardless of if it is a multiple of 5 presses (CU-j6y98q).
 - Start new session if "current" session is waiting for reply and it has been >= 2 hours since last update (CU-j6yfct).
+- Ability to have multiple fallback phone numbers (CU-pv8hd5).
 
 ## [3.4.0] - 2020-03-29
+
 ### Added
 - Add CSV button to dashboard (CU-c8htxp).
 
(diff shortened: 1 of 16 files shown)


trisapeace:
Task linked: [CU-pv8hd5 allow multiple fallback numbers](https://app.clickup.com/t/pv8hd5)

CHANGELOG.md
```diff
@@ -10,61 +11,85 @@ the code was deployed.
 
 ## [Unreleased]
 
+### Added
+
+- Ability to have multiple fallback phone numbers (CU-pv8hd5).
+
 ## [3.3.0] - 2020-03-08
+
```
trisapeace, line 19: Note: All these extra lines were added in by prettier.

CHANGELOG.md
```diff
@@ -194,31 +253,40 @@ the code was deployed.
...
 - pi install script uses config file instead of interactive input.
 
 ## [1.0-pi-heartbeat] - 2019-01-11
+
 ### Added
+
 - Initial pi.
 
-[Unreleased]: https://github.com/bravetechnologycoop/BraveButtons/compare/v3.3.0...HEAD
+[unreleased]: https://github.com/bravetechnologycoop/BraveButtons/compare/v3.3.0...HEAD
```
trisapeace, line 289: Note: This case change was done by prettier. The link still seems to work even with the different case.

server/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "server.js",
   "dependencies": {
     "body-parser": "^1.18.3",
-    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#v2.4.0",
+    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#td-multiple-fallbacks",
```
trisapeace, line 8: Remember to update this once https://github.com/bravetechnologycoop/brave-alert-lib/pull/10 is merged.

server/test/serverTest.js
```diff
@@ -556,7 +556,7 @@ describe('Chatbot server', () => {
     })
 
     it('should send the initial text message after a valid single click request to /flic_button_press', async () => {
-      // eslint-disable-next-line prettier/prettier
+      // prettier-ignore
```
risatronic, line 559: I'm curious about this change because the reason I did it using the `eslint-disable` is that `prettier-ignore` seems to affect chunks of formatting without a way to reliably target a single line, when I had used it it got rid of a bunch of formatting complaints in the same area that hadn't been fixed yet and should have been instead of just this `await` chain

risatronic (approved):
Just had a small question about eslint vs prettier, but everything looks good to me! :)

server/test/serverTest.js
trisapeace: Ooo, good to know. I'll have to look into it more. I switched it because my VSCode extension was still formatting this line even though it had the eslint ignore. That's probably because I have the prettier plugin turned on but not the eslint one (I find it's underlines while I'm still typing really annoying). I can switch it back.

risatronic: oooh weird! Maybe mine is extension-related too then but I think I have both on 🤔

trisapeace: I switched it back. Although I thiiiiiink the prettier ignore would have worked too, according to its docs (https://prettier.io/docs/en/ignore.html#javascript), but I'm happy with the eslint ignore as long as I know that that's what's happening.

trisapeace:
To knock out a few niggling chore-type Tasks, I also included changes for:
* https://app.clickup.com/t/p107rc (rename `registry` table to `buttons`)
* https://app.clickup.com/t/307j4h (enumerate where our logs are in the `README`)
* Fixed instructions in `README` on how to make encrypted variables in Travis.

CHANGELOG.md
```diff
@@ -196,31 +260,41 @@ the code was deployed.
...
 
 ## [1.0-pi-heartbeat] - 2019-01-11
+
 ### Added
+
 - Initial pi.
 
-[Unreleased]: https://github.com/bravetechnologycoop/BraveButtons/compare/v3.3.0...HEAD
+[unreleased]: https://github.com/bravetechnologycoop/BraveButtons/compare/v3.4.0...HEAD
+[3.4.0]: https://github.com/bravetechnologycoop/BraveButtons/compare/v3.3.0...v3.4.0
```
trisapeace, line 297: This link was missed in the last deployment.

mariocimet:
looks great :) also I suspect this is a record pull request-to-task ratio

mariocimet (approved):

trisapeace:
Haha, thanks. Yeah, I normally like to keep it to one issue per pull request, but they seemed small, I was on a roll, and I wanted fewer Tasks in our Backlog :smile:

schwarrrtz (approved):
this looks good, and the tests all pass for me locally. thanks for knocking off a few extra tasks as well.

PR #91, opened 2021-04-19
Always update the numPresses of the most recent unresponded session (CU-tb9pvc)

trisapeace:
- This fixes a bug that only appears in our last release because we
  now allow multiple sessions to be left in "Waiting for Reply".
  This ensures that it will only get the most recent of those sessions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ the code was deployed.
 
 ## [Unreleased]
 
+### Fixes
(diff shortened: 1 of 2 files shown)


trisapeace:
Task linked: [CU-tb9pvc Fix problem where multiple sessions are started](https://app.clickup.com/t/tb9pvc)

mariocimet (approved):

schwarrrtz (approved):
this looks good, all the tests pass for me locally, and it should match the previous behaviour.

PR #98, opened 2021-05-13
Update Brave Alert Lib

trisapeace:
- To change `logSentry` to a message instead of an error
- To fix the route in handleTwilioMessage error messages
- ~~Didn't update the CHANGELOG because these two changes are modifications of things that already have an entry in the upcoming release's CHANGELOG.~~

Tested this by
- Deploying on `chatbot-dev`
- Replied to a Button-press alert with a phone that wasn't the Responder phone, and got the error message `21|BraveServer  | 2021-05-13 23:23:25.853: Bad request to /alert/sms: Invalid Phone Number`
- Manually changed the values in the `hubs` table to simulate a Hub reconnecting and disconnecting. This made a Sentry info message that looks like this: https://sentry.io/organizations/brave-technology-coop/issues/2397959388/?project=5743498&query=&statsPeriod=14d

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -5,7 +5,7 @@
   "main": "server.js",
   "dependencies": {
     "body-parser": "^1.18.3",
-    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#v3.2.0",
+    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#v3.3.0",
     "cookie-parser": "^1.4.3",
     "express": "^4.16.4",
     "express-session": "^1.15.6",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ the code was deployed.
 
 ## Unreleased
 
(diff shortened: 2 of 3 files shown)


server/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "server.js",
   "dependencies": {
     "body-parser": "^1.18.3",
-    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#v3.2.0",
+    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#td-fix-api-error-messages",
```
trisapeace, line 8: Note to self, update this after https://github.com/bravetechnologycoop/brave-alert-lib/pull/14 is merged.

schwarrrtz (approved):
This looks good, and it sounds like you've tested it thoroughly. Once the corresponding PR is merged in brave-alert-lib and you've updated the `package.json` I'd say this is good to go.

server/package.json

PR #102, opened 2021-06-04
Add GET /alert/historicAlerts endpoint and `sessions.responded_at` column (CU-hjwfx2)

trisapeace:
- To be used by the Alert App
- Added `responded_at` column to the `sessions` table because this
  needs to be returned by the new endpoint
- Added functionality to set the `session.responded_at` column at
  the appropriate times
- Added "Responded At" column to the Chatbot Dashboard
- Moved `SESSION_RESET_TIMEOUT` into the .env so that it would
  be easily accessible from `BraveAlerterConfigurator.js` without
  needing to pass it in to the constructor. This is also good for
  manual testing because waiting 2 hours can suck. So it'll be good
  to be able to just change the .env and redeploy to wait for a
  shorter period of time to test this functionality
- Added and changed tests accordingly
- Changed all sinon.stub(helpers, 'log*') calls to use sinon.spy
  instead so that I could still see the log output in the console
  while the spy tracks when and with what parameters it is called

Corresponding brave-alert-lib PR: https://github.com/bravetechnologycoop/brave-alert-lib/pull/15

**Test plan:**
- :heavy_check_mark: Deploy Buttons to `chatbot-dev` and go through all the flows with a real Button watching when and if the `responded_at` value changed
- :heavy_check_mark: Hit `GET https://chatbot-dev.brave.coop/alert/historicAlerts` for my test Button and see that
   - :heavy_check_mark: Not-completed Button presses in the last 2 hours are not returned
   - :heavy_check_mark: Completed Button presses in the last 2 hours are returned
   - :heavy_check_mark: All Button presses longer ago than 2 hours are returned
   - :heavy_check_mark: Different values for `X-API-KEY` (missing, empty string, no matching API key, matching API key)
   - :heavy_check_mark: Different values for `maxHistoricAlerts` (missing, null, empty string, 0, -1, 1, 99999)
   - :heavy_check_mark: Sessions that return HistoricAlerts where `alertType` is either `BUTTON_URGENT` or `BUTTONS_NOT_URGENT` depending on the value of `numPresses`
   - :heavy_check_mark: HistoricAlerts which are returned have the values I'd expect based on the Sessions they were created from
- :heavy_check_mark: Hit the `GET https://chatbot-dev.brave.coop/alert/historicAlerts` endpoint from the Alert App and render its results
- :heavy_check_mark: Look at Chatbot Dashboard to see the new Responded At column and that it has the right value in it
- :heavy_check_mark: Change `SESSION_RESET_TIMEOUT` and see it work as expected for new Button presses and for determining what is a Historic Alert

### DIFF
--- a/server/BraveAlerterConfigurator.js
+++ b/server/BraveAlerterConfigurator.js
@@ -1,5 +1,5 @@
 /* eslint-disable class-methods-use-this */
-const { BraveAlerter, AlertSession, ALERT_STATE, helpers, Location, SYSTEM } = require('brave-alert-lib')
+const { BraveAlerter, AlertSession, ALERT_TYPE, ALERT_STATE, helpers, Location, SYSTEM, HistoricAlert } = require('brave-alert-lib')
 const db = require('./db/db.js')
 
 class BraveAlerterConfigurator {
@@ -9,6 +9,7 @@ class BraveAlerterConfigurator {
       this.getAlertSessionByPhoneNumber.bind(this),
       this.alertSessionChangedCallback,
       this.getLocationByAlertApiKey.bind(this),
+      this.getHistoricAlertsByAlertApiKey.bind(this),
       true,
       this.getReturnMessage.bind(this),
     )
@@ -103,6 +104,10 @@ class BraveAlerterConfigurator {
           session.fallBackAlertTwilioStatus = alertSession.fallbackReturnMessage
         }
 
+        if (alertSession.alertState === ALERT_STATE.WAITING_FOR_CATEGORY) {
+          session.respondedAt = await db.getCurrentTime(client)
+        }
+
         await db.saveSession(session, client)
       }
 
@@ -128,6 +133,24 @@ class BraveAlerterConfigurator {
     // Even if there is more than one matching installation, we only return one and it will
     // be used by the Alert App to indentify this installation
     return new Location(installations[0].name, SYSTEM.BUTTONS)
+  }
+
+  createHistoricAlertFromRow(row) {
+    const alertType = row.num_presses > 1 ? ALERT_TYPE.BUTTONS_URGENT : ALERT_TYPE.BUTTONS_NOT_URGENT
+    return new HistoricAlert(row.id, row.unit, row.incident_type, alertType, row.num_presses, row.created_at, row.responded_at)
+  }
+
+  // Historic Alerts are those with status "Completed" or that were last updated longer ago than the SESSION_RESET_TIMEOUT
+  async getHistoricAlertsByAlertApiKey(alertApiKey, maxHistoricAlerts) {
+    const maxTimeAgoInMillis = helpers.getEnvVar('SESSION_RESET_TIMEOUT')
+
+    const historicAlerts = await db.getHistoricAlertsByAlertApiKey(alertApiKey, maxHistoricAlerts, maxTimeAgoInMillis)
+
+    if (!Array.isArray(historicAlerts)) {
+      return null
+    }
+
+    return historicAlerts.map(this.createHistoricAlertFromRow)
   }
 
   getReturnMessage(fromAlertState, toAlertState, incidentCategories) {
--- a/server/db/014-addrespondedattosessions.sql
+++ b/server/db/014-addrespondedattosessions.sql
@@ -1,0 +1,21 @@
+DO $migration$
+    DECLARE migrationId INT;
+    DECLARE lastSuccessfulMigrationId INT;
+BEGIN
+    -- The migration ID of this file
+    migrationId := 14;
+
+    -- Get the migration ID of the last file to be successfully run
+    SELECT MAX(id) INTO lastSuccessfulMigrationId
+    FROM migrations;
+
+    -- Only execute this script if its migration ID is next after the last successful migration ID
+    IF migrationId - lastSuccessfulMigrationId = 1 THEN
+        ALTER TABLE sessions 
+        ADD COLUMN responded_at timestamptz;
+
+        -- Update the migration ID of the last file to be successfully run to the migration ID of this file
+        INSERT INTO migrations (id)
+        VALUES (migrationId);
+    END IF;
+END $migration$;
--- a/server/db/db.js
+++ b/server/db/db.js
@@ -22,7 +22,7 @@ types.setTypeParser(types.builtins.NUMERIC, value => parseFloat(value))
 
 function createSessionFromRow(r) {
   // prettier-ignore
-  return new SessionState(r.id, r.installation_id, r.button_id, r.unit, r.phone_number, r.state, r.num_presses, r.created_at, r.updated_at, r.incident_type, r.notes, r.fallback_alert_twilio_status, r.button_battery_level)
+  return new SessionState(r.id, r.installation_id, r.button_id, r.unit, r.phone_number, r.state, r.num_presses, r.created_at, r.updated_at, r.incident_type, r.notes, r.fallback_alert_twilio_status, r.button_battery_level, r.responded_at)
 }
 
 function createInstallationFromRow(r) {
@@ -264,7 +264,7 @@ async function getAllSessions(clientParam) {
   return null
 }
 
-async function createSession(installationId, buttonId, unit, phoneNumber, numPresses, buttonBatteryLevel, clientParam) {
+async function createSession(installationId, buttonId, unit, phoneNumber, numPresses, buttonBatteryLevel, respondedAt, clientParam) {
   let client = clientParam
   const transactionMode = typeof client !== 'undefined'
 
@@ -273,9 +273,9 @@ async function createSession(installationId, buttonId, unit, phoneNumber, numPre
       client = await pool.connect()
     }
 
-    const values = [installationId, buttonId, unit, phoneNumber, ALERT_STATE.STARTED, numPresses, buttonBatteryLevel]
+    const values = [installationId, buttonId, unit, phoneNumber, ALERT_STATE.STARTED, numPresses, buttonBatteryLevel, respondedAt]
     const { rows } = await client.query(
-      'INSERT INTO sessions (installation_id, button_id, unit, phone_number, state, num_presses, button_battery_level) VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING *',
+      'INSERT INTO sessions (installation_id, button_id, unit, phone_number, state, num_presses, button_battery_level, responded_at) VALUES ($1, $2, $3, $4, $5, $6, $7, $8) RETURNING *',
       values,
     )
 
@@ -322,10 +322,11 @@ async function saveSession(session, clientParam) {
       session.notes,
       session.fallBackAlertTwilioStatus,
       session.buttonBatteryLevel,
+      session.respondedAt,
       session.id,
     ]
     await client.query(
-      'UPDATE sessions SET installation_id = $1, button_id = $2, unit = $3, phone_number = $4, state = $5, num_presses = $6, incident_type = $7, notes = $8, fallback_alert_twilio_status =$9, button_battery_level=$10 WHERE id = $11',
+      'UPDATE sessions SET installation_id = $1, button_id = $2, unit = $3, phone_number = $4, state = $5, num_presses = $6, incident_type = $7, notes = $8, fallback_alert_twilio_status = $9, button_battery_level = $10, responded_at = $11 WHERE id = $12',
       values,
     )
   } catch (e) {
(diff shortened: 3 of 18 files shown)


trisapeace:
Task linked: [CU-hjwfx2 Frontend: Alert History List View: Implementation](https://app.clickup.com/t/hjwfx2)

server/BraveAlerterConfigurator.js
```diff
@@ -103,6 +104,10 @@ class BraveAlerterConfigurator {
           session.fallBackAlertTwilioStatus = alertSession.fallbackReturnMessage
         }
 
+        if (!session.respondedAt && session.state !== ALERT_STATE.STARTED && session.state !== ALERT_STATE.WAITING_FOR_REPLY) {
+          session.respondedAt = await db.getCurrentTime(client)
+        }
+
```
trisapeace, line 110: I'm not super happy with this `if`-statement. It looks like the other ones above it, but because it depends on `session.state` which could have changed in an earlier `if`-statement, it's sort of different. The value of `session.state` at this point should be the new State, not the old one, whereas up until this point, I'd been using old values from `session` and new values from `alertSession`. However, I can't use `alertSession` here because sometimes it's not sent because it's not being updated (for example, when we update the Twilio Fallback Status, we don't change the Session State). Although I think we've agreed that we can remove the Twilio Fallback Status column and no longer update it, but this doesn't seem to be the time to do that.

server/db/db.js
```diff
@@ -636,6 +637,34 @@ async function getInstallationWithSessionId(sessionId, clientParam) {
...
+  let client = clientParam
+  const transactionMode = typeof client !== 'undefined'
+
+  try {
+    if (!transactionMode) {
+      client = await pool.connect()
+    }
+
+    // prettier-ignore
+    const { rows } = await client.query('SELECT s.id, b.unit, s.incident_type, s.num_presses, s.created_at, s.responded_at FROM sessions AS s LEFT JOIN buttons AS b ON s.button_id = b.button_id LEFT JOIN installations AS i ON s.installation_id = i.id WHERE i.alert_api_key = $1 AND (s.state = $2 OR s.updated_at < now() - $3 * INTERVAL \'1 millisecond\') ORDER BY s.created_at DESC LIMIT $4', [alertApiKey, ALERT_STATE.COMPLETED, maxTimeAgoInMillis, maxHistoricAlerts])
```
trisapeace, line 650: I think this is our most complicated SQL query so far in this code base. But don't worry, I wrote integration tests for it. Let me know what you think.

server/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "server.js",
   "dependencies": {
     "body-parser": "^1.18.3",
-    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#v3.3.0",
+    "brave-alert-lib": "https://github.com/bravetechnologycoop/brave-alert-lib#td-getHistoricAlerts",
```
trisapeace, line 8: Reminder to self to update this one https://github.com/bravetechnologycoop/brave-alert-lib/pull/15 is merged.

server/server.js
```diff
@@ -19,7 +19,6 @@ const db = require('./db/db.js')
 const FLIC_THRESHOLD_MILLIS = 210 * 1000
 const HEARTBEAT_THRESHOLD_MILLIS = 75 * 1000
 const PING_THRESHOLD_MILLIS = 320 * 1000
-const SESSION_RESET_TIMEOUT = 2 * 60 * 60 * 1000
```
trisapeace, line 22: I already added this value to the `.env` on the `chatbot-dev` and `chatbot` servers.

server/db/014-addrespondedattosessions.sql
```diff
@@ -0,0 +1,21 @@
...
+
+    -- Get the migration ID of the last file to be successfully run
+    SELECT MAX(id) INTO lastSuccessfulMigrationId
+    FROM migrations;
+
+    -- Only execute this script if its migration ID is next after the last successful migration ID
+    IF migrationId - lastSuccessfulMigrationId = 1 THEN
+        ALTER TABLE sessions 
+        ADD COLUMN responded_at timestamptz;
+
```
trisapeace, line 16: What do you think about us setting the `responded_at` value to the `updated_at` value for all existing rows in `sessions` (after first temporarily turning off the updated_at trigger for this table)? Or maybe only for the rows which aren't 'Started' or 'Waiting for reply'. We'd sort of be lying because it won't necessarily be the time when the Button press was first responded to, but in most cases, I think it would be close.

server/db/db.js
schwarrrtz: @trisapeace that would be awesome, thanks.

server/BraveAlerterConfigurator.js
```diff
@@ -130,6 +135,24 @@ class BraveAlerterConfigurator {
     return new Location(installations[0].name, SYSTEM.BUTTONS)
   }
 
+  createHistoricAlertFromRow(row) {
+    const alertType = row.num_presses > 1 ? ALERT_TYPE.BUTTONS_URGENT : ALERT_TYPE.BUTTONS_NOT_URGENT
+    return new HistoricAlert(row.id, row.unit, row.incident_type, alertType, row.num_presses, row.created_at, row.responded_at)
+  }
+
+  // Historic Alerts are those with status "Completed" or that were last updated longer ago than the SESSION_RESET_TIMEOUT
```
schwarrrtz, line 143: This comment, or something like it, would be helpful in `db.js` as well to explain what the complicated Postgres query is doing.

server/BraveAlerterConfigurator.js
trisapeace: I was trying to cover too many edge cases, let's change this to:
```
if (alertSession.alertState !== ALERT_STATE.WAITING_FOR_CATEGORY) {          
   session.respondedAt = await db.getCurrentTime(client)        
}
```

schwarrrtz: that looks great, it's much easier to follow now.

server/BraveAlerterConfigurator.js
trisapeace: Done!

server/db/db.js
```diff
@@ -636,6 +637,49 @@ async function getInstallationWithSessionId(sessionId, clientParam) {
...
+      LEFT JOIN buttons AS b ON s.button_id = b.button_id
+      LEFT JOIN installations AS i ON s.installation_id = i.id
+      WHERE i.alert_api_key = $1
+      AND (
+        s.state = $2
+        OR s.updated_at < now() - $3 * INTERVAL '1 millisecond'
+      )
+      ORDER BY s.created_at DESC
+      LIMIT $4
+      `,
```
trisapeace, line 663: What do you two think of this for multi-line string. It looks much more readable to me.

server/BraveAlerterConfigurator.js
trisapeace: Fixed. Thanks for working through this one with me @schwarrrtz :smile:

server/db/db.js
trisapeace: Done.

server/BraveAlerterConfigurator.js
schwarrrtz: thanks!

server/db/014-addrespondedattosessions.sql
schwarrrtz: I don't think this is necessary, since `responded_at` will need to accommodate null values anyhow in order to accommodate unresponded sessions. also, we may want to calculate something like average response time in the future, and backfilling `responded_at` in this way might introduce some outliers that could skew the result of calculations like this.

if I remember correctly, @trisapeace you were talking about the old sessions rendering in a misleading way if `responded_at` is null for sessions that actually got a response, but maybe the fronted can handle that by looking at the session state?

schwarrrtz: actually, we could backfill `responded_at` for rows where state is 'Waiting for category', since those are the rows where the last update was the initial response. this assumes that no other batch operations or maintenance have affected the `updated_at` value.

schwarrrtz: but I think I would still prefer to leave the old sessions with blank `responded_at` values for now - we can always backfill them later if it becomes a problem.

trisapeace: Yes, I was concerned that for Sessions that obviously had been responded to (like they had an incident category assigned to them), would display "Alert response: No response" in the expanded part of the Historic Alert view like the bottom view in this screenshot.
image (9)

Yes, we could work in some logic in the Alert App to not do this, but I'm thinking that by the time the Alert App is live, all of the Sessions that would be returned will have a Response Time.

server/db/db.js
schwarrrtz: this is much better!

schwarrrtz: also the comment makes it easier to understand what the query is doing. thanks.

server/test/integration/dbTest/getHistoricAlertsByAlertApiKeyTest.js
```diff
@@ -0,0 +1,296 @@
...
+    })
+  })
+
+  describe('if there is one matching session', () => {
+    beforeEach(async () => {
+      await db.clearSessions()
+      await db.clearButtons()
+      await db.clearInstallations()
+
+      // Insert a single installation with a two buttons and more than maxHistoricAlerts sessions
```
schwarrrtz, line 75: where does `maxHistoricAlerts` get set? it looks like only one session is created here - this comment doesn't seem to match that?

trisapeace: You're right. Comment is copied from elsewhere and isn't correct. I'll fix it.

trisapeace: But to answer question about `maxHistoricAlerts`, it's sent as a parameter to `getHistoricAlertsByAlertApiKey` in the test.

schwarrrtz: ah yes, I see that. thanks!

server/db/014-addrespondedattosessions.sql
schwarrrtz: makes sense.

server/test/unit/BraveAlerterConfiguratorTest/getHistoricAlertsByAlertApiKeyTest.js
```diff
@@ -0,0 +1,148 @@
...
+      id: 'id',
+      unit: 'unit',
+      incident_type: 'incidentType',
+      num_presses: 2,
+      created_at: new Date('2020-01-20T10:10:10.000Z'),
+      responded_at: new Date('2020-01-20T10:12:40.000Z'),
+    }
+    sinon.stub(db, 'getHistoricAlertsByAlertApiKey').returns([results])
+
+    const historicAlerts = await this.braveAlerter.getHistoricAlertsByAlertApiKey('alertApiKey', 'maxHistoricAlerts')
```
schwarrrtz, line 39: I don't understand why the string `'maxHistoricAlerts'` is passed to `getHistoricAlertsByAlertApiKey` throughout this test file. shouldn't `maxHistoricAlerts` be an integer?

schwarrrtz: ah ok, it's because of the stub.

trisapeace: Just because it's not actually being used. This is a unit test for `getHistoricAlertsByAlertApiKey`, but all it does with its second parameter is pass it along to `db.getHistoricAlertsByAlertApiKey` which is stubbed out above to return the `results` array no matter what it's passed.

So I could pass it an integer, but it also wouldn't affect the test in any way. What do you think would be clearer from reading the test?

server/BraveAlerterConfigurator.js
schwarrrtz: no worries! :)

schwarrrtz (approved):
other than the minor changes to the comments I've asked about, I'd say this is good to go. it sounds like you've tested it thoroughly, and written a bunch of new automated tests as well - nice.

server/test/unit/BraveAlerterConfiguratorTest/getHistoricAlertsByAlertApiKeyTest.js
schwarrrtz: I think if I had read the tests more closely I would have seen the stub and realized what was going on sooner. I did figure it out eventually.

you could make a `const maxHistoricAlerts = 'maxHistoricAlerts'` with a comment explaining what's going on, then reuse the constant throughout the tests - that would make it super clear - but I don't think it's necessary.

server/test/integration/dbTest/getHistoricAlertsByAlertApiKeyTest.js
trisapeace: Fixed.

server/test/unit/BraveAlerterConfiguratorTest/getHistoricAlertsByAlertApiKeyTest.js
trisapeace: OK, I'll leave it. Sorry about the confusion.

mariocimet (approved):

server/package.json

PR #103, opened 2021-06-11
Fix PM2 issues after Ubuntu 20.04 upgrade

trisapeace:
- After upgrading to Ubuntu 20.04, the version of pm2 that we use
  changed from 3.5.1 to 4.5.6
- For some reason, this meant that running the `setup_server.sh` as
  it was and also the scheduled `pm2 restart` cron job failed to
  properly restart the job. Instead, they would do weird things like
  spawning a new server.js node process but not kill the old one, so
  the server would still be running the old code. Other times, it
  would leave the `pm2 status` view with no running code. Very strange
- These changes will run PM2 and the BraveServer node app as the brave
  user instead of root. This means that when running pm2 commands
  manually, we no longer need to use sudo. And when running pm2 command
  in setup_server.sh, we need to specify to run them using the brave
  user

My definition of "a working state":
1. When I run `pm2 status`, I see BraveServer and pm2-logrotate running:
image
2. When I run `ps aux | grep 'node'`, I see only one node process running `server.js`:  
   ```
   brave@buttons-server-dev:~$ ps aux | grep 'node'
   brave      18446  0.2  4.1 794600 42000 ?        Ssl  22:44   0:01 node /home/brave/.pm2/modules/pm2- 
   logrotate/node_modules/pm2-log
   brave      18515  0.8  7.8 822852 78448 ?        Ssl  22:44   0:04 node /home/brave/BraveButtons/server/server.js
   brave      18654  0.0  0.0   8160   736 pts/0    S+   22:54   0:00 grep --color=auto node
   ```
3. When I run `ps aux | grep 'pm2', I see only one PM2 God daemon and it's running with PM2_HOME 
= /home/brave/.pm2:
   ```
   brave@buttons-server-dev:~$ ps aux | grep 'pm2'
   brave      16138  0.2  4.6 798212 46936 ?        Ssl  22:41   0:02 PM2 v5.1.0: God Daemon (/home/brave/.pm2)
   brave      18446  0.2  4.2 794600 42468 ?        Ssl  22:44   0:01 node /home/brave/.pm2/modules/pm2-l 
   logrotate/node_modules/pm2-log
   brave      18652  0.0  0.2   8160  2456 pts/0    S+   22:54   0:00 grep --color=auto pm2
   ```
4. Changes made to `heartbeatDashboard.mst` since the previous deployment attempt appear in https://chatbot-dev.brave.coop/heartbeatDashboard (this represents to me that any new code changes would have been deployed)
5. Pushing my real test button, I can make it through the normal chatbot flow and watch the Chatbot Dashboard update

I tested this by:
- Given I'm already in a working state. When I run `sudo ./setup_server.sh .env` . Then I am in a working state again.
- Given I'm already in a working state. When the cron job runs. Then I am in a working state again.

### DIFF
--- a/server/setup_server.sh
+++ b/server/setup_server.sh
@@ -3,10 +3,9 @@
 set -e
 original_dir=$(pwd)
 cd $(dirname "$0")
-export HOME=$(bash <<< "echo ~$SUDO_USER")
 
 if [[ $EUID > 0 ]]; then
-    echo "this script needs sudo privelages to run correctly."
+    echo "this script needs sudo privileges to run correctly."
     cd $original_dir
     exit 1
 elif [[ ! -n "$1" ]]; then
@@ -26,7 +25,7 @@ else
         fi
     done < $1
 
-    echo "please enter two IP addresses to whitelist for SSH (separated by a space):"
+    echo "please enter two IP addresses to allowlist for SSH (separated by a space):"
     read firstIP secondIP
 
     ufw default deny incoming
@@ -43,6 +42,8 @@ else
     apt install -y nodejs npm certbot postgresql postgresql-contrib
     npm install -g pm2 n
     n 12.18.3         # keep this in sync with .nvmrc for Travis
+    PATH=$PATH        # needed to set the new path for this version of node
+    setcap cap_net_bind_service=+ep /usr/local/bin/node   # allows non-root to use port 443
     npm ci
 
     echo "Please enter in order the name and responder phone number and one fallback phone number for the first installation (separated by a space):" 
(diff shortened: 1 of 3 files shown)


server/setup_server.sh
```diff
@@ -51,23 +52,38 @@ else
...
+    mkdir -p /var/log/brave
+    chmod 777 /var/log/brave
+    touch /var/log/brave/pm2-crontab.log
+    chmod 666 /var/log/brave/pm2-crontab.log
 
     # restart server weekly to ensure it uses the latest certificates (certbot renews them automatically)
     echo "
-    PATH=/bin:/usr/bin:/usr/local/bin
-    @weekly env HOME=$HOME pm2 restart BraveServer
+    @weekly /usr/sbin/runuser -u brave -- /usr/local/bin/pm2 restart BraveServer >> /var/log/brave/pm2-crontab.log 2>&1
```
trisapeace, line 75: I'm going to check on Monday whether this ran over the weekend.

trisapeace (commented):
This same code (but rebased off of the `td-getHistoricAlerts` branch) is up on the new `chatbot-dev` right now if you'd like to play around with this.

server/setup_server.sh
trisapeace: It ran successfully over the weekend and is back to a working state.

mariocimet (approved):

server/setup_server.sh
```diff
@@ -26,7 +25,7 @@ else
         fi
     done < $1
 
-    echo "please enter two IP addresses to whitelist for SSH (separated by a space):"
+    echo "please enter two IP addresses to allowlist for SSH (separated by a space):"
```
schwarrrtz, line 28: nice, good catch.

schwarrrtz (approved):
this looks good, we went over the bulk of the changes during the code review meeting on Friday, and it sounds like you've tested it thoroughly.

PR #105, opened 2021-06-14
Add VSCode launch config for debugging Mocha tests for /server

trisapeace:
- So that those of us who use VS Code will be able to use it to
  debug Mocha tests without any extra work

Some screenshots of how it works from the equivalent change in BraveSensors: https://github.com/bravetechnologycoop/BraveSensor-Server/pull/114/files


schwarrrtz (approved):
looks good, it only touches a VSCode-specific file and the changelog, so I'd say this is good to merge. maybe I should try VSCode...

PR #106, opened 2021-07-02
Fix time formatting in CSV Export (CU-zkkach)

trisapeace:
- Postgres doesn't use the standard date/time format string that is
  used in programming languages. Instead, "HH" means 12-hour hours
  when what we wanted was "HH24" to get 24-hour hours and "MM" means
  months when what we wanted was "MI" to get minutes

Test Plan:
- :heavy_check_mark:  Deploy to `chatbot-dev`, export the CSV, check all 5 date/time columns against the values in the DB and make sure that they are right even when the hour is > 12 and the minute is different than the month's number.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ the code was deployed.
 
 - Updated deployment instructions (CU-wb8719).
 
+### Fixed
+
+- Time strings in CSV Export now have the correct hour and minute values (CU-zkkach).
+
 ## [4.0.0] - 2021-06-15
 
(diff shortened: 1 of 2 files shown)


trisapeace:
Task linked: [CU-zkkach Fix hour and minute formatting in CSV Export](https://app.clickup.com/t/zkkach)

schwarrrtz (approved):
Sounds like you've tested this thoroughly. Thanks for reformatting the query string as well, it looks much better!

mariocimet (approved):

PR #107, opened 2021-07-08
added audit-ci to Travis to check for security issues in server deps

schwarrrtz:
This pull request adds the same `audit-ci` tool to Travis as I added in this BraveAlert pull request: https://github.com/bravetechnologycoop/brave-alert-app/pull/11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,8 +68,9 @@ matrix:
       before_script:
         - bash setup_postgresql_local_dev.sh
 
-      # run linter and unit tests
+      # run security audit, linter and unit tests
       script:
+        - if [ "${TRAVIS_PULL_REQUEST}" != "false" ]; then npx audit-ci --config ./audit-ci.json; fi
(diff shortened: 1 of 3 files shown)


trisapeace (approved):
Oh sweet, so this one already had no high security issues, that's really good news. I guess Dependabot was keeping us on our toes. Thanks!

trisapeace:
@schwarrrtz Could you please add the same thing to Sensors while you're at it?

schwarrrtz:
@trisapeace I haven't checked Sensors, but brave-alert-lib has audit issues that need to be fixed - I want to move on to other things instead of testing those changes, but I think it would be good to add `audit-ci` to all our Node repos eventually.

trisapeace:
OK interesting. Thanks! I agree, it would be worth adding. How about adding a ClickUp Task for it for those other two repos?

mariocimet (approved):

PR #111, opened 2021-07-20
POST /flic_button_press returns 401 when given invalid API Key (CU-gwxnde)

trisapeace:
- This makes it more difficult for people to hit our endpoint
  pretending to be a Button when really they aren't.
- Gives us better security against this attack
- Also made the formatting and prettier-ignores more consistent in `serverTest.js`

## Test Plan
- :heavy_check_mark: Check that there have been no `INVALID` logs in production in all the logs that we have right now (right now, this covers July 6-20)
- :heavy_check_mark: Deploy to chatbot-dev
- :heavy_check_mark: Configure my real buttons to hit `POST /flic_button_press` with the real API key
   - :heavy_check_mark: Starts chatbot session
- :heavy_check_mark: Configure my real buttons to hit `POST /flic_button_press` with an invalid API key
   - :heavy_check_mark: Doesn't starts chatbot session
   - :heavy_check_mark: Logs the invalid API attempt
- :heavy_check_mark: Configure my real buttons to hit `POST /flic_button_press` with no API key
   - :heavy_check_mark: Doesn't starts chatbot session
   - :heavy_check_mark: Logs the invalid API attempt
- :heavy_check_mark: Use Postman to hit `POST /flic_button_press` with an invalid API key and with no API key to see that it returns 401

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ the code was deployed.
 ### Changed
 
 - Updated deployment instructions (CU-wb8719).
+- `POST /flic_button_press` returns 401 Unauthorized when given an invalid API key (CU-gwxnde).
 
 ### Fixed
 
--- a/server/server.js
+++ b/server/server.js
@@ -314,11 +314,10 @@ app.post('/flic_button_press', Validator.header(['button-serial-number']).exists
       const apiKey = req.query.apikey
 
       // Log the vaiditiy of the API key
-      // TODO (CU-gwxnde) Replace this with a 401 unauthorized if invalid
       if (apiKey !== helpers.getEnvVar('FLIC_BUTTON_PRESS_API_KEY')) {
         helpers.logError(`INVALID api key from '${buttonName}' (${serialNumber})`)
-      } else {
-        helpers.log(`VALID api key from '${buttonName}' (${serialNumber})`)
+        res.status(401).send()
+        return
       }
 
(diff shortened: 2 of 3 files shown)


trisapeace:
Task linked: [CU-gwxnde Chatbot sessions should not be able to be started unless by one of our Buttons](https://app.clickup.com/t/gwxnde)

mariocimet:
looks great! I noticed that we are quite a bit more detailed in our use of 40X error codes in Sensors than Buttons (except for Particle where everything is a 200), opting mostly for a generic 400 except for this new situation with an API key.

is it worth it to add this detail to Buttons? do the consumers of the error codes care?

approved

mariocimet (approved):

trisapeace:
It's a good question. In this particular case (`/flic_button_press`), our consumer is the Flic App, which I don't think cares at all and I don't even think that we can see a history of those requests and responses.

As for in general, I can see a good case for better error messages from the endpoints used by the Alert App, but those are all covered within `brave-alert-lib`.

The `/heartbeat/*` ones are all ones that we would hit manually and we'd be the only users of, so they are probably fine.

The `/heartbeat` one is being sent by the RPis. I'm not sure off the top of my head what it does with it, but as long as the error message is on one side or the other, then we'll be able to pick it up.

As for the ones for the dashboard, that would be nice if they were better. I think we'll have some more dashboard changing coming up as the DB schema starts to change. We can try to remember to do it then!

PR #119, opened 2021-10-05
No longer update Responded At when given an invalid incident category (CU-1ae1j60)

trisapeace:
This is just a one-liner, but I took the opportunity to do some clean up in the session-related functions in `db.js` since they will be tested in this test plan and are sort of related to what I changed.

## Test plan
- :heavy_check_mark: Deploy to `chatbot-dev`
- :heavy_check_mark: Use Postman to simulate a button press and see the respondedAt value is `null`
   - :heavy_check_mark: Respond 'OK' and see the respondedAt value change to the current time on the Dashboard
   - :heavy_check_mark: Respond to the incident category question with a valid incident category and see the respondedAt value does not change
   - :heavy_check_mark: Respond to the incident category question with an invalid incident category and see the respondedAt value does not change
   - :heavy_check_mark: Respond to the incident details question and see the respondedAt value does not change
   - :heavy_check_mark: Do not respond 'OK' to the button press and see that the respondedAt value remains `null` after the reminder message and after the fallback message

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ the code was deployed.
 - Updated `ALERT_STATE` to `CHATBOT_STATE`.
 - `POST /alert/designatedevice` also logs the given Responder Push ID (CU-10xfkhr).
 
+### Fixed
+
+- No longer updates a session's Responded At time if given an invalid incident category (CU-1ae1j60).
+
 ## [4.1.0] - 2021-07-26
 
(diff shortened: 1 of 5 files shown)


trisapeace:
Task linked: [CU-1ae1j60 Fix bug where `responded_at` is updated when the user gives an invalid incident category](https://app.clickup.com/t/1ae1j60)

PR #122, opened 2021-10-16
Added firmware for range testing a single RB50 button CU-1hym8ja

mariocimet:
Tested on a Boron with Device OS 2.2.0 and 2.3.0-rc. The firmware uses the device name and type to look for RB50s so you don't need to hardcode a single test buttons address. I noted some stability problems while using it, namely frequent errors like this:

ERROR: Failed to load session data from persistent storage

but unsure if they're related to the single Boron I was testing on. 

In order to handle multiple buttons, the main change that needs to happen is for us to store a hash table of button addresses and control bytes, and to do the checks in checkAndPublishButtons() with that table instead of a single "previousControlByte".

image

I would prefer to use a short name filter style strategy to scan instead of a MAX_LENGTH for the scan and then checking the name. This is only available for Particle OS 3+ and a bit out of scope

### DIFF
--- a/particle-button-hub/project.properties
+++ b/particle-button-hub/project.properties
@@ -1,0 +1,2 @@
+name=boron-rb50-button-hub
+dependencies.CircularBuffer=1.3.3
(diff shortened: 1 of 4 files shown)


elektraDak (approved):
LGTM (from knowledge I have)

trisapeace:
Closing without merging because it looks like we will not be using the RB50s.

PR #124, opened 2021-11-01
Allow releaseinfo changes on RPis (CU-1kxmhta)

trisapeace:
- Raspian OS version has changed from Buster to something else. So now,
  Buster, the one that is on our devices, is considered `oldstable`
  instead of `stable`. This fix came from:
  https://forums.raspberrypi.com/viewtopic.php?t=318302

# Test Plan
- :heavy_check_mark: Run `apt update --allow-releaseinfo-change` manually on the Kye7e hub just to make sure that it works with `apt update` (when we did this during the last deployment, we used `apt-get update`)
- :heavy_check_mark: Make these change to `setup_pi.sh` file on Alvis 1R Hub (which was missed during the previous deployment because it was offline at the time) and manually run `sudo ./BraveButtons/pi/setup_pi.sh ./BraveButtons/pi/pi_config.ini` to make sure that the changes to the script will work
- :heavy_check_mark: Run the Ansible script on just Alvis 1R to make sure that it will work for subsequent deployment (`ansible-playbook -i ~/BraveButtonsConfig/ansible/inventory-prod.yaml  ~/BraveButtons/ops/update_pi_fleet.yaml --ask-vault-pass --limit alvis_pi_1R`)

PR #127, opened 2021-11-12
setup_pi.yaml no longer triggers apt update using the apt module

schwarrrtz:
ansible doesn't support allowing releaseinfo changes; apt update still runs eventually because of the setup script.

### DIFF
--- a/ops/setup_pi.yaml
+++ b/ops/setup_pi.yaml
@@ -40,12 +40,13 @@
         name: "OPENSSH_PUBLIC_KEY_HOLDER"
         public_key: "{{ OPENSSH_KEYPAIR_RETURN.public_key }}"
 
+    # don't use update_cache here since there is no way to use --allow-releaseinfo-change in ansible
+    # unless they merge this: https://github.com/ansible/ansible/pull/62824
     - name: install git
       become: yes
       apt:
         name: git
         state: present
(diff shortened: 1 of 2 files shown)


schwarrrtz:
this came up during programming of the Norland hub. initially the ansible playbook was failing because it was using `apt` to install git and it was trying to run `apt update` without `--allow-releaseinfo-change`. @trisapeace fixed this same issue in the main setup script in https://github.com/bravetechnologycoop/BraveButtons/pull/124. however, since ansible doesn't have a way to run apt with this flag I just removed the option that triggers the update. this means we might install a slightly older version of git, which should be ok.

after making this change, the Norland hub programming worked normally.

ops/setup_pi.yaml
```diff
@@ -40,12 +40,13 @@
         name: "OPENSSH_PUBLIC_KEY_HOLDER"
         public_key: "{{ OPENSSH_KEYPAIR_RETURN.public_key }}"
 
+    # don't use update_cache here since there is no way to use --allow-releaseinfo-change in ansible
+    # unless they merge this: https://github.com/ansible/ansible/pull/62824
```
trisapeace, line 44: Good comment. Thanks.

trisapeace (approved):

PR #129, opened 2021-11-18
Update darkstat parsing (CU-1k0zhbg)

trisapeace:
@schwarrrtz What sort of testing should I do on this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ the code was deployed.
 ### Changed
 
 - `setup_pi.yaml` no longer triggers `apt update` using the apt module since ansible doesn't support allowing releaseinfo changes.
-- Improved darkstat parsing to look at all entries (CU-1k0zhbg).
+- Improved darkstat parsing to look at all entries, to skip null IP addresses, and convert day(s) to seconds (CU-1k0zhbg).
 
 ## [6.0.0] - 2021-11-29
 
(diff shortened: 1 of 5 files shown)


trisapeace:
Task linked: [CU-1k0zhbg review darkstat parsing code](https://app.clickup.com/t/1k0zhbg)

trisapeace:
Task linked: [CU-40u01q handle multiple instances of the Flic Hub MAC when parsing darkstat output](https://app.clickup.com/t/40u01q)

schwarrrtz:
I just ran this branch on the Norland 1 hub and it seems to be working equivalently to the current production release (v5.0.0) which makes sense, since these fixes only handle edge cases.

I'd say this is probably good to deploy to production - perhaps we could deploy this only to one client initially, as long as the other changes in the next production release allow for this.

schwarrrtz (approved):

trisapeace:
Thanks so much for running that. I'm glad to hear that it works as expected in the normal case!

Good idea. I've scheduled the next Buttons deployment for Monday. How about I don't merge this one until after that deployment? Then we can do a deployment with just this to one hub (probably either Viv 2 or Viv 3 because they're local and often have the issue where Last Seen (Ping) is high and the other values are low).

schwarrrtz:
@trisapeace sounds like a good plan.

PR #132, opened 2021-11-27
Change master branch name to match preferred branching scheme (CU-mn5q4g)

trisapeace:
- Renamed `master` --> `main`
- Updated README and CHANGELOG accordingly
- Documentation here: https://app.clickup.com/2434616/v/dc/2a9hr-2261/2a9hr-4347
- Also fixed a 404 link in packages.json that had `master` in the URL

Once this pull request is approved, I will *not* merge it into `master`. Instead, I will 
1. Double-check that this branch is up-to-date with `master` (if not, then I'll rebase with `master`)
2. Update GitHub's default branch to be `main`
3. Change the target of all open pull requests from `master` to `main`
4. Delete the `master` branch

Once that is done, then the steps that I suggest everyone does on their local are:
1. Delete the `master` branch
2. Pull the `main` branch

PR #133, opened 2021-11-30
setup_pi.sh sets the hostname in a more robust way (CU-40tzeq)

schwarrrtz:
this pull request changes `setup_pi.sh` so that the hostname can be repeatedly changed. the previous implementation would only work when changing the hostname from the default value of 'raspberrypi'.

thanks @trisapeace for the idea on how to implement this.

I tested this by manually running `setup_pi.sh` on a test hub while changing the hostname in `pi_config.ini`. each time the hostname was updated properly in both `/etc/hostname` and `/etc/hosts`. the hub was also able to connect to the internet and appeared normally on the heartbeat dashboard.

### DIFF
--- a/pi/setup_pi.sh
+++ b/pi/setup_pi.sh
@@ -96,8 +96,9 @@ else
   chmod 700 /usr/local/brave
 
   # change the hostname according to the config
+  old_hostname=$(</etc/hostname)
   hosts_file=$(</etc/hosts)
-  hosts_file="${hosts_file//raspberrypi/$hostname}"
+  hosts_file="${hosts_file//$old_hostname/$hostname}"
   echo "$hosts_file" > /etc/hosts
   echo "$hostname" > /etc/hostname
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ the code was deployed.
 
 ## [Unreleased]
 
+
+### Changed
+
+- `setup_pi.sh` sets the hostname in a more robust way (CU-40tzeq).
+
 ## [6.1.0] - 2021-11-29
 
 ### Changed


trisapeace:
Task linked: [CU-40tzeq fix hostname error when sudoing on RPi](https://app.clickup.com/t/40tzeq)

schwarrrtz:
@trisapeace one of the tests is failing unexpectedly on Travis - the same one as in your recent PR (#132)

trisapeace (approved):
This looks great. Thanks for testing it so thoroughly. So we'll still have to do the Viv ones manually one time, but then this will make it super awesome in the future. I'm glad that it worked!

Here's a pull request with a change that I'm pretty sure will make that test stop being flaky: https://github.com/bravetechnologycoop/BraveButtons/pull/134

pi/setup_pi.sh
```diff
@@ -96,8 +96,9 @@ else
   chmod 700 /usr/local/brave
 
   # change the hostname according to the config
+  old_hostname=$(</etc/hostname)
   hosts_file=$(</etc/hosts)
-  hosts_file="${hosts_file//raspberrypi/$hostname}"
+  hosts_file="${hosts_file//$old_hostname/$hostname}"
```
trisapeace, line 101: Hmmm, actually, one more question about this one. Is `raspberrypi` already in the `/etc/hostname` file initially?

trisapeace: Just found this article, which makes it looks like yes, it is in `/etc/hostname` initially.
https://www.tomshardware.com/how-to/raspberry-pi-change-hostname

schwarrrtz:
had to rebase and force push since the auto-merge of the CHANGELOG would have made it seem like this had already been released.

PR #134, opened 2021-11-30
Fix flaky getActiveAlertsByAlertApiKey integration test

trisapeace:
- The `db.getHistoricAlertsByAlertApiKey` function compares
`maxTimeAgoInMillis` against the `updated_at` value in the DB. It's
possible that this test is running so fast that it has been less than
one millisecond since the `db.createSession` completed, in which case
we would get the wrong value back. So I changed the DB query to compare
<= instead of <. This will not make any real difference to the behaviour
in real life (because a 1ms difference doesn't really matter to humans)
but should make this test more reliable.


schwarrrtz (approved):
nice!

PR #139, opened 2021-12-31
Changed `pm2 stop BraveServer` to `pm2 kill` in deployment instructions

trisapeace:
- Changed `pm2 stop BraveServer` to `pm2 kill` to handle problem that
  I ran into where environment variables were cached and didn't update
  by a simple re-deploy

While I was in there, I made these other changes that will help me during production deployments:
- Moved production deployment instructions to top of README so that it's easier to find given that it's the part of the README that we use the most often
- Moved section of how to deploy to the Raspberry Pi fleet using
  Ansible directly into the production deployment instructions to
  eliminate the need to jump around the document during deployment
- Changed the big Ansible command to deploy the whole fleet to use
  the `inventory-prod.yaml` instead of a placeholder to put either
  the production or dev inventory. This will reduce issues during
  deployments and should be relatively obviously that it would need
  to change if doing a rare dev deployment
  
Are you OK with all this @sampathsatti ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,75 @@
 # BraveButtons [![Build Status](https://app.travis-ci.com/bravetechnologycoop/BraveButtons.svg?branch=main)](https://app.travis-ci.com/bravetechnologycoop/BraveButtons)
 
+# How to release a new production version & update the entire Button system (server and RPi fleet)
+
+The following instructions apply if you are deploying a new version of the code, optionally including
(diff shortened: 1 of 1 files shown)


sampathsatti:
This looks great Theresa, I didn't know pm2 cached environment variables like that through a re-deploy. Great catch!

Appreciate the readme changes as well, they should greatly increase the quality and usability.

PR #164, opened 2022-06-07
Stops texts from completing older sessions (CU-q37bbk)

trisapeace:
- This was an annoying bug that's been around for awhile. If the
  most recent session was Completed and the server received a new
  text from the Responder Phone, it would treat it as though it
  were a response to the most recent incompleted session.
- We are about to implement multiple Responder phones, which will
  make it much more likely that a text message will be sent after
  a session is complete, thus making this bug more likely to be
  seen

## Test Plan
- :heavy_check_mark: Deploy to Dev
- :heavy_check_mark: Run through all the normal cases and check status on Dashboard
   - :heavy_check_mark: Respond OK right away, see incident category question
   - :heavy_check_mark: Respond valid incident category, see details question
   - :heavy_check_mark: Respond invalid incident category, see invalid category message
   - :heavy_check_mark: Respond with notes, see thank you message
   - :heavy_check_mark: Respond after the thank you message (do this more than once)
      - :heavy_check_mark: See no active session message
      - :heavy_check_mark: No Sentry log
   - :heavy_check_mark: Do not respond OK right away
      - :heavy_check_mark: See Reminder Message
   - :heavy_check_mark: Respond OK after the Reminder Message
      - :heavy_check_mark: See same flow as if you pushed OK right away
   - :heavy_check_mark: Do not respond OK after the Reminder Message
      - :heavy_check_mark: See Fallback message 
   - :heavy_check_mark: Multiple Button presses
      - :heavy_check_mark: See Urgent message

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ the code was deployed.
 
 - If the Last Uplink for a gateway is undefined, return `null` (CU-2dm6j58).
 
+### Fixed
+
+- Text messages sent to the chatbot only interact with the most recent session (CU-q37bbk).
+
 ## [8.2.0] - 2022-06-02
 
 ### Changed
--- a/server/BraveAlerterConfigurator.js
+++ b/server/BraveAlerterConfigurator.js
@@ -68,7 +68,7 @@ class BraveAlerterConfigurator {
     let alertSession = null
 
     try {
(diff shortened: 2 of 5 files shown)


trisapeace:
Task linked: [CU-q37bbk chatbot should only process SMS responses for the most recent session](https://app.clickup.com/t/q37bbk)

PR #176, opened 2022-08-16
Send disconnection messages to Sentry for RAK Buttons (CU-208arfp)

trisapeace:
- For now, only sends to Sentry so that we can restrict
  it to just us Brave people until we know for sure that these messages
  are reliable and not too noisy

## Test Plan
- :heavy_check_mark: Run through normal chatbot
- :heavy_check_mark: Set my client and my Button to active
   -  :heavy_check_mark: Sentry receives the Button disconnect message when buttonVitals.createdAt is older than the threshold
   - :heavy_check_mark: Sentry receives the Button reconnection message when buttonVitals.createdAt is more recent than the threshold
- :heavy_check_mark: Set my client to active but Button to inactive
   - :heavy_check_mark: Do not receive any Sentry messages
- :heavy_check_mark: Set my client and Button to inactive
   - :heavy_check_mark: Do not receive any Sentry messages
- :heavy_check_mark: Set my client to inactive but Button to active
   - :heavy_check_mark: Do not receive any Sentry messages

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ the code was deployed.
 
 - Disconnection/reconnection SMS messages to Heartbeat Phone Numbers for RAK Gateways (CU-208arfp).
 - Low battery SMS messages to Heartbeat Phone Numbers for RAK Buttons (CU-208arfp).
+- Disconnection/reconnection Sentry messages for RAK Buttons (CU-208arfp).`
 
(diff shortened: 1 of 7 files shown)


trisapeace:
Task linked: [CU-208arfp Implement automatic downtime and low battery Sentry messages for RAK Buttons](https://app.clickup.com/t/208arfp)

PR #177, opened 2022-10-13
Pressing RAK Button 4 starts an alert (CU-2uadjfh)

trisapeace:
- The new injection-moulded Buttons will push on Button 4 instead of Button 3. We still have 3D-printed Buttons in production, so did not remove the alerting ability of Button 3, which it uses.

## Test plan
- ✔️ Deploy to Dev
- ✔️ Push Button 4 once and multiple times, should start an alert and also send expected urgent alerts
- ✔️ Push Button 3 once and multiple times, should start an alert and also send expected urgent alerts
- ✔️ Push Button 1 once and multiple times, should NOT start any alerts
- ✔️ Push Button 2 once and multiple times, should NOT start any alerts
- ✔️ Push a combination of Buttons 3 and 4, should start an alert and also send expected urgent alerts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ the code was deployed.
 
 ## [Unreleased]
 
+### Added
+
+- RAK Button 4 can be used to trigger alerts (CU-2uadjfh).
+
 ## [10.2.0] - 2022-08-22
 
 ### Changed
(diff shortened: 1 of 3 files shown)


trisapeace:
Task linked: [CU-2uadjfh RAK Button 4 presses should also trigger an alert](https://app.clickup.com/t/2uadjfh)

PR #180, opened 2022-11-29
Send Gateway vitals and RAK low battery alerts to Responder Phones (CU-3bft2mp)

trisapeace:
- We have found that these alerts are not too noisy, so we should get them in the hands of our Responders as soon as possible so they are apprised of the current status of their system

## Test Plan
- Deploy to dev
- Generate a RAK Button low battery message and see that it is sent to the Responder Phones and Heartbeat Phones
- Generate a RAK Gateway disconnection message and see that it is sent to the Responder Phones and Heartbeat Phones
- Generate a RAK Gateway reconnection message and see that it is sent to the Responder Phones and Heartbeat Phones

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ the code was deployed.
 ### Added
 
 - Spanish messages (CU-38tu7z8).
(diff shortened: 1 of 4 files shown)


trisapeace:
Task linked: [CU-3bft2mp Send Buttons Gateway disconnection/reconnection and RAK Button low battery messages to the Responder Phone as well](https://app.clickup.com/t/3bft2mp)

PR #202, opened 2023-09-27
Update README

trisapeace:
- So production deployment instructions look more similar to the ones in Sensors
- Also updated the `npm start` command so that it will actually work


README.md
```diff
@@ -50,19 +44,29 @@ making a new tag during step 2. This is essentially redeploying an older version
         - To: The phone number to send the final confirmation text to (i.e. your phone number)
         - From: Any number in the format +12223334444 (we don't use this number for anything, but Twilio will complain if it's missing or incorrectly formatted)
 
+## 4. Deply the server on Production
```
tmackinnon, line 47: Small typo with "Deploy"

README.md
```diff
@@ -78,8 +82,14 @@ making a new tag during step 2. This is essentially redeploying an older version
         - To: The phone number to send the final confirmation text to (i.e. your phone number)
         - From: Any number in the format +12223334444 (we don't use this number for anything, but Twilio will complain if it's missing or incorrectly formatted)
 
+## 8. Celebrate
```
tmackinnon, line 85: Love that you put this in!

README.md
```diff
@@ -78,8 +82,14 @@ making a new tag during step 2. This is essentially redeploying an older version
...
+
 1. send a message to the `#buttons-aa-general` Slack channel letting everyone know that the deployment is finished and list the changes in this deployment from the `CHANGELOG`
 
+1. update the ClickUp Tasks
+
+1. if appropriate, send a Feature Change Announcement email to the clients
```
tmackinnon, line 91: Out of curiosity - is there a template/example for this type of email that in the future we should refer to?

tmackinnon (commented):
Looks great! This added more clarity too, which is much appreciated. Thanks!

README.md
trisapeace: Thanks!

README.md
trisapeace: You had a good point that it was missing :smile:

README.md
trisapeace: I've sent some out using HubSpot to Sensors clients before. I don't remember the last one that I did for a Buttons change. But in either case, there is a template that Dana created that I used for these, but after the first one I just copied the previous and made modifications. I'll forward one of the previous ones to you all so you can see what they have looked like in the past.

Basically, if the change that we make is one that our clients should or might care about, then we like to show off and tell them about it. Dana will know how to do this and she might also post stuff on our social media about it.

trisapeace:
Note that Travis is failing right now due to a security issue. The fix for it is in this PR: https://github.com/bravetechnologycoop/BraveButtons/pull/203

README.md
tmackinnon: Thanks Theresa! This is really good go know going forward.

README.md
```diff
@@ -1,42 +1,36 @@
...
    1. push the new version to GitHub: `git push origin main --tags`
 
    1. update the `production` branch: `git checkout production && git merge main && git push origin production`
 
-1. on your local machine, in the `buttons-config` repository:
-
-   1. login to AWS SSO: `aws sso login --profile brave-management`
-
-   1. ensure you have the latest version of the code: `git pull origin main`
+## 2. Update the Environment Variables (if necessary)
```
ahmaddaadaa, line 27: I am also considering adding the following step:

If you need to install AWS CLI on your machine, please follow the instructions provided in this link: [AWS CLI Installation Guide](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html).

If necessary, follow these steps:

-  In your root directory, create a `.aws` folder.

- Inside the `.aws` folder, use the `nano` text editor to create or edit the `config` file. Paste the `[profile-brave-management]` key configuration into this file.

- If you encounter an issue where running `aws --version` results in `-bash: [route]: no such file or directory`, follow these additional steps:

   - Open the `.bashrc` file using the `nano` text editor.
   - Scroll to the bottom of the file and ensure that the path for your scripting is correctly defined. For example: `export PATH="$PATH:$HOME/.rvn/bin:[route]"`.
   - Save the changes.

-  After saving the `.bashrc` file, run `source .bashrc` to apply the changes.

-  Finally, run `aws --version` again to confirm that everything is functioning correctly.

These steps should help you set up the necessary AWS configurations and ensure that the AWS CLI operates without issues.

ahmaddaadaa (changes requested):

README.md
trisapeace: Hi Ahmad. The environment variables for Buttons live in the `.env` file on the `chatbot.brave.coop` server (or `chatbot-dev.brave.coop in the case of the Development environment).

We don't use AWS for Buttons. So I won't be adding your suggestion here. However, please take a look at this PR about updates to the Sensors README and see if there is a good place to add your suggestion there: https://github.com/bravetechnologycoop/BraveSensor/pull/227

README.md
```diff
@@ -1,42 +1,36 @@
...
+1. send a message to the `#buttons-aa-general` Slack channel letting everyone know that you are doing a deployment and to expect some downtime.
+
 1. on your local machine, in the `BraveButtons` repository:
 
    1. pull the latest code ready for release: `git checkout main && git pull origin main`
 
    1. decide on an appropriate version number for the new version
 
-   1. update CHANGELOG.md by moving everything in `Unreleased` to a section for the new version
+   1. update CHANGELOG.md by moving everything in `Unreleased` to a section for the new version. Remember to scroll down to the bottom to update the links!
```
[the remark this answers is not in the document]
trisapeace: Good suggestion! I added some description of exactly what to do. Please let me know if this gives you what you need? Thanks!

micahdbak (approved):
All looks good, instructions are clear and verbose!

ahmaddaadaa (commented):
Hi Therese,
Nice work! However, it appears that you are encountering Travis error, likely due to the recent update in the Travis Babel version. Please make sure your local version is in sync with ours by running:
`npm install @babel/traverse@7.23.2 --save` push your updates again and it should pass

trisapeace:
Good point @ahmaddaadaa . I just committed a change to fix the security issue. 

Your solution to use `npm install @babel/traverse@7.23.2 --save` to fix the problem does successfully make the security issue go away. However, our code does not explicitly use `@babel/traverse`. Instead, it is further down in our dependency tree. 

In general, it is good to only include libraries that our code uses in our `package.json`. This is because it's possible that in the future, the library that depended on @babel/traverse changes and no longer uses @babel/traverse anymore. In this case, we would want our code to no longer use @babel/traverse either, but if it's included in our `package.json` then we will include it whether it's being used or not.

Therefore, instead of adding `@babel/traverse` to the `package.json` as you suggested using the command `npm install @babel/traverse@7.23.2 --save`, I used `npm audit fix` to update just the `package-lock.json` file.

Check it out and let me know what you think.

ahmaddaadaa (approved):
Perfect! Thanks Theresa

PR #205, opened 2023-10-11
POST /api/message-clients (CU-w9bcb5)

micahdbak:
The API call /api/message-clients sends text messages to active clients.

It works as follows:

For all clients that are sending vitals, sending alerts, and have at least one button that is sending vitals and sending alerts, send a Twilio message containing the message that was POSTed to /api/message-clients.
Requires proper submission of the Particle Accelerator API key (braveKey); this is because the API call is intended to be called from PA.

Upon successful submission of the API key, the server will respond with a JSON object containing information about contacted/failed clients:

```json
{
    "status": "success",
    "message": "This is the text message that was sent!",
    "contacted": [
        {
            "to": "+11111111111",
            "from": "+12222222222",
            "clientId": "01234567-89ab-cdef-0123-456789abcdef",
            "clientDisplayName": "Test Client"
        },
        {
            "to": "+13333333333",
            "from": "+14444444444",
            "clientId": "abcdef01-2345-6789-abcd-ef0123456789",
            "clientDisplayName": "Test Client"
        }
    ],
    "failed": []
}
```

Where `contacted` and `failed` are arrays of objects containing `to`, `from`, `clientId`, and `clientDisplayName` members.
All objects in `contacted` were successfully contacted; Twilio gave the status of "queued".
All objects in `failed` weren't successfully contacted; Twilio rejected the request, and did not queue the message.
Requests can fail for a variety of reasons, for example, an incorrect `from` phone number.

This response information is intended to be displayed in the PA dashboard.

Additionally, upon failure to send a Twilio message, the following message will be logged internally; i.e., visible from the CloudWatch dashboard in AWS:

```
Failed to send Twilio message: {"to":"+11111111111","from":"+12222222222","clientId":"01234567-89ab-cdef-0123-456789abcdef","clientDisplayName":"Test Client"}.
```

Test plan:
- [x] write integration tests to make sure /api/message-clients parses active clients properly
  - [x] for a request that uses the correct PA API key, and where Twilio is operating correctly
    - [x] should respond with status 200 (OK)
    - [x] should respond with the same message that was posted
    - [x] should attempt to message all active clients
    - [x] should not attempt to message any non-active clients
    - [x] should respond with contacted clients in the `contacted` field of the response body
    - [x] should respond with no clients in the `failed` field of the response body
  - [x] for a request that uses the correct PA API key, and where Twilio is not operating correctly
    - [x] should respond with status 200 (OK)
    - [x] should respond with the same message that was posted
    - [x] should attempt to message all active clients
    - [x] should not attempt to message any non-active clients
    - [x] should respond with no clients in the `contacted` field of the response body
    - [x] should respond with active clients in the `failed` field of the response body
  - [x] for a request that uses the incorrect PA API key
    - [x] should respond with status 401 (Unauthorized)
    - [x] should not message anyone
  - [x] for a request that does not provide any PA API key
    - [x] should respond with status 401 (Unauthorized)
    - [x] should not message anyone
- [x] create several clients and buttons on local database; test that this API call decides which clients to message properly
- [x] use this API call on the development server and prove that it works properly with real Twilio API keys

### DIFF
--- a/server/db/db.js
+++ b/server/db/db.js
@@ -142,6 +142,37 @@ async function getClients(pgClient) {
   return []
 }
 
+async function getActiveClients(pgClient) {
+  try {
+    const results = await helpers.runQuery(
+      'getActiveClients',
+      `
+        SELECT c.*
+	FROM clients c
+	INNER JOIN (
+	  SELECT DISTINCT client_id AS id
+	  FROM buttons
+	  WHERE is_sending_alerts AND is_sending_vitals
+	) AS b
+	ON c.id = b.id
+	WHERE c.is_sending_alerts AND c.is_sending_vitals
+	ORDER BY c.display_name;
+      `,
+      [],
+      pool,
+      pgClient,
+    )
+
+    if (results === undefined) {
+      return null
+    }
+
+    return await Promise.all(results.rows.map(r => createClientFromRow(r)))
+  } catch (err) {
+    helpers.logError(err.toString())
+  }
+}
+
 async function getButtons(pgClient) {
   try {
     const results = await helpers.runQuery(
@@ -1290,6 +1321,7 @@ module.exports = {
   getDataForExport,
   getHistoricAlertsByAlertApiKey,
   getClients,
+  getActiveClients,
   getClientsWithAlertApiKey,
   getClientWithId,
   getClientWithSessionId,
--- a/server/routes.js
+++ b/server/routes.js
@@ -1,6 +1,3 @@
-// Third-party dependencies
-const express = require('express')
-
 // In-house dependencies
 const { clickUpHelpers } = require('brave-alert-lib')
 const dashboard = require('./dashboard')
@@ -19,12 +16,7 @@ function configureRoutes(app) {
   app.get('/vitals', dashboard.sessionChecker, dashboard.renderVitalsPage)
 
   app.post('/login', dashboard.submitLogin)
-  app.post(
-    '/pa/aws-device-registration',
-    pa.validateAwsDeviceRegistration,
-    clickUpHelpers.clickUpChecker,
-    pa.handleAwsDeviceRegistration,
-  )
+  app.post('/pa/aws-device-registration', pa.validateAwsDeviceRegistration, clickUpHelpers.clickUpChecker, pa.handleAwsDeviceRegistration)
   app.post('/pa/buttons-twilio-number', pa.validateButtonsTwilioNumber, clickUpHelpers.clickUpChecker, pa.handleButtonsTwilioNumber)
   app.post('/rak_button_press', rak.validateButtonPress, rak.handleButtonPress)
 
--- a/server/routes.js
+++ b/server/routes.js
@@ -8,8 +8,6 @@ const pa = require('./pa')
 const rak = require('./rak')
 const api = require('./api')
 
-const jsonBodyParser = express.json()
-
 function configureRoutes(app) {
   app.get('/', dashboard.sessionChecker, dashboard.redirectToHomePage)
   app.get('/dashboard', dashboard.sessionChecker, dashboard.renderDashboardPage)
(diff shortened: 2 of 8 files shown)


trisapeace:
Task linked: [CU-w9bcb5 Add a way to send a custom message to all active responder and system status phones in Buttons and Sensors](https://app.clickup.com/t/w9bcb5)

tmackinnon (approved):
This is looking great Micah! And good job fixing the files so they are similar to BraveSensor - this will be helpful for whenever the time comes to merge the two backends. Since this is so similar to what you did in BraveSensor I have no changes, so approved from me.

ahmaddaadaa (approved):
Nice work Micah!

server/db/db.js
```diff
@@ -142,6 +142,37 @@ async function getClients(pgClient) {
...
+        SELECT c.*
+	FROM clients c
+	INNER JOIN (
+	  SELECT DISTINCT client_id AS id
+	  FROM buttons
+	  WHERE is_sending_alerts AND is_sending_vitals
+	) AS b
+	ON c.id = b.id
+	WHERE c.is_sending_alerts AND c.is_sending_vitals
+	ORDER BY c.display_name;
```
trisapeace, line 159: Do we get any benefit from ordering the results?

server/routes.js
```diff
@@ -1,13 +1,9 @@
-// Third-party dependencies
-const express = require('express')
-
 // In-house dependencies
 const { clickUpHelpers } = require('brave-alert-lib')
 const dashboard = require('./dashboard')
 const pa = require('./pa')
 const rak = require('./rak')
-
-const jsonBodyParser = express.json()
```
trisapeace, line 10: Oh cool! You were able to get rid of `jsonBodyParser`. Yay!

server/test/integration/apiTest/messageClientsTest.js
```diff
@@ -0,0 +1,328 @@
...
+
+chai.use(chaiHttp)
+chai.use(sinonChai)
+
+const expect = chai.expect
+const sandbox = sinon.createSandbox()
+const braveKey = helpers.getEnvVar('PA_API_KEY_PRIMARY')
+const badBraveKey = 'badbravekey'
+const message = 'Hello, world!'
+const testFromPhoneNumber = '+11234567890'
```
trisapeace, line 21: We can also have different fromPhoneNumbers for each client.

server/test/integration/apiTest/messageClientsTest.js
```diff
@@ -0,0 +1,328 @@
...
+        clientId: activeClients[i].id,
+        displayName: `Button for ${activeClients[i].displayName}`,
+        buttonSerialNumber: `Button${activeClients[i].id.substr(0, 8)}`,
+        isSendingAlerts: true,
+        isSendingVitals: true,
+      })
+    }
+
+    // non-active clients
+    const nonActiveClients = []
```
trisapeace, line 100: How about active Clients who do not have any active Buttons?

server/test/integration/apiTest/messageClientsTest.js
```diff
@@ -0,0 +1,328 @@
...
+      expect(response.body.data).to.not.be.undefined
+      expect(response.body.data.message).to.be.a('string').that.equals(message)
+    })
+
+    it('should attempt to message all active clients', async () => {
+      await this.agent.post('/api/message-clients').send({ braveKey, message })
+
+      testExpectPhoneNumbers.forEach(pn => {
+        expect(twilioHelpers.sendTwilioMessage).to.be.calledWith(pn, testFromPhoneNumber, message)
+      })
```
trisapeace, line 201: Ooooo, interesting one.

Generally speaking, a single test should have a single assertion/expects-statement in it (although there are always exceptions). This is because you don't want to hide the results of later tests if an earlier test fails.

In this case, where the array of `testExpectPhoneNumbers` is `['+11111111111', '+12222222222', '+13333333333', '+14444444444']`, if this test failed on `+11111111111` then we don't know whether it would have failed for all the phone numbers in the array or only some of them. Which could be useful information for fixing the failure.

I think there is a function like `getCalls()` that can be used with stubs like `twilioHelpers.sendTwilioMessage.getCalls()` to get a list or array of all the times it was called (documentation: https://sinonjs.org/releases/v17/spy-call/) and then then you could compare that with the calls that you were expecting and provide a more detailed failure for this kind of test.

Up to you if you want to go back and fiddle with this since this code is already merged. But I'm putting it here for future reference!

server/test/integration/apiTest/messageClientsTest.js
```diff
@@ -0,0 +1,328 @@
...
+      // stub twilioHelpers.sendTwilioMessage
+      sandbox.stub(twilioHelpers, 'sendTwilioMessage').returns(undefined)
+    })
+
+    afterEach(async () => {
+      sandbox.restore()
+    })
+
+    it('should respond with status 200 (OK)', async () => {
+      const response = await this.agent.post('/api/message-clients?test=1').send({ braveKey, message })
```
trisapeace, line 238: What is the `test=1` query parameter?

server/test/integration/apiTest/messageClientsTest.js
```diff
@@ -0,0 +1,328 @@
...
+
+const expect = chai.expect
+const sandbox = sinon.createSandbox()
+const braveKey = helpers.getEnvVar('PA_API_KEY_PRIMARY')
+const badBraveKey = 'badbravekey'
+const message = 'Hello, world!'
+const testFromPhoneNumber = '+11234567890'
+
+// information about clients that should be contacted
+const testExpectPhoneNumbers = ['+11111111111', '+12222222222', '+13333333333', '+14444444444']
```
trisapeace, line 24: This is great. Pretty complex test case that you had to set up in order to do this. Thanks for sticking with it. 

I've added some comments on other cases that could be included. When it starts getting picky like that, then you might think about moving some of this to the unit level and calling `api.messageClients` directly with a mocked `req`, `res` and `db`.

trisapeace (commented):
Awesome.

PR #207, opened 2023-10-26
Log client messages for disconnected and reconnected RAK Buttons (CU-860raw0d7)

tmackinnon:
Adding functionality for when RAK buttons disconnect and reconnect. Currently the functionality of checking a buttons disconnection/reconnection status already exists and it currently logs button info upon disconnection/reconnection. If we were to send a twilio notification for every button disconnection/reconnection as it stands now, the number of messages would potentially overwhelm clients. 

This PR adds some extra safeguards to limit the number of alerts. To test if these changes have an impact, it also includes a log that we can monitor for a couple weeks before adding the notification functionality. After this implementation, we will have a better idea of the number of notifications that could be sent per client, and then can determine if we can to replace the test logs with twilio notifications. 

## Overall Plan: 

**Phase 1**
- [x] Add a client message object that keeps tracks of all disconnected buttons and reconnected buttons per client
- [x] Add functionality that checks if gateway is offline - if a client has at least one gateway offline, don't send message to client
- [x] Once all disconnection/reconnection info is gathered, log one message per client noting the affected buttons

**Phase 2**
- Monitor logs implemented above
- Compare the # of newly added logs (from Phase 1, that happen once per client to the existing disconnection/reconnection logs (the ones that happen once per button)
- If new logs < existing logs & would not overwhelm clients then move to Phase 3

**Phase 3 - New PR/Task**
- Add send notification message functionality either in place of or with the new logs 
- Add feature flag table in DB so we can turn feature on/off per client (in case of too many messages)

### checkButtonHeartbeatTest - add to existing tests

**Client is sending vitals & Button is sending vitals:**
- [x] When a button has disconnected, log disconnection message, log client message
- [x] When button has reconnected, log reconnection message, log client message
- [x] When button remains connected, it shouldn't log any Sentry messages

**Client is sending vitals & Button is not sending vitals:** 
- [x] It shouldn't log any Sentry messages

**Client is not sending vitals & Button is sending vitals:**
- [x]  It shouldn't log any Sentry messages

**Client is not sending vital & Button is not sending vitals**
- [x] It shouldn't log any Sentry messages

### checkButtonHeartbeatTest - create new tests
**Client is sending vitals & Buttons are sending vitals**
- [x] When both buttons (from the same client) disconnect: Should log a disconnection message for each button, should log one client message
- [x] When the same client has a disconnected button, a reconnected button and a button that has remained connected (no change): Should log a disconnection message, should log a reconnection message, should log one client message, and shouldn't log anything for the button that remained connected
- [x] When the same client has two buttons where one has just disconnected and one remains disconnected: Should log one disconnection message, should log one client message (not including the button that remains disconnected)
- [X] When client has a disconnected button & a disconnected gateway: should log the disconnection, should not log the client message

### Manual Testing in Dev Environment:
Manipulate Database for following test cases:
- [x] When a button has disconnected, disconnection message and client message are logged to Sentry
- [x] When button has reconnected, reconnection message and client message are logged to Sentry
- [x] When button remains connected, it shouldn't log any Sentry messages
- [x] When 2 buttons (from the same client) disconnect: Should log a disconnection message for each button and one client message on Sentry
- [x] When the same client has a disconnected button, a reconnected button: Should log a disconnection message, should log a reconnection message, should log one client message
- [x] When the same client has two buttons where one has just disconnected and one remains disconnected: Should log one disconnection message and one client message (not including the button that remains disconnected) on Sentry
- [x] When client has a disconnected button & a disconnected gateway: should log the disconnection, should not log the client message

### Other Testing
- [x] Travis passed


server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
```diff
@@ -3,13 +3,16 @@ const { expect, use } = require('chai')
...
 const { DateTime } = require('luxon')
 
 // In-house dependencies
 const { helpers, factories } = require('brave-alert-lib')
 const db = require('../../../db/db')
 const { buttonFactory, buttonsVitalFactory } = require('../../testingHelpers')
-const vitals = require('../../../vitals')
+require('../../mocks/tMock')
+
+const vitals = rewire('../../../vitals')
```
trisapeace, line 15: Cool. I'm glad you figured out the `rewire`!

In this particular case, since the code is ours, you could also have exported the `sendNotifications` function from `vitals.js`. But I like that you used `rewire` because that way we didn't have to do that export!

server/vitals.js
```diff
@@ -142,6 +142,10 @@ async function checkButtonBatteries() {
 
 async function checkButtonHeartbeat() {
   try {
+    // Objets to track disconnected and reconnected buttons
+    const disconnectedButtons = {}
```
trisapeace, line 146: Cool. I think you're on a good track.

server/vitals.js
```diff
@@ -140,8 +140,32 @@ async function checkButtonBatteries() {
...
+        const buttonNames = disconnectedButtons.join(', ')
+        buttonLogMessage += ` The following buttons have been disconnected: ${buttonNames}.`
+      }
+
+      if (reconnectedButtons.length > 0) {
+        const buttonNames = reconnectedButtons.join(', ')
+        buttonLogMessage += ` The following buttons have been reconnected: ${buttonNames}.`
+      }
+
+      helpers.logSentry(`Button status change for: ${clientDisplayName}.${buttonLogMessage}`)
```
trisapeace, line 159: This looks good to me. 

Will these button names be in any particular order?

server/vitals.js
```diff
@@ -170,10 +207,23 @@ async function checkButtonHeartbeat() {
...
+          // Store the client info
+          if (!clientMessages[client.id]) {
+            clientMessages[client.id] = {
+              client,
+              disconnectedButtons: [],
+              reconnectedButtons: [],
+            }
+          }
+          // Store the reconnected button name
+          clientMessages[client.id].reconnectedButtons.push(button.displayName)
```
trisapeace, line 219: Cool. The use of a dictionary/object looked weird to me to begin with, but now I see that you needed to add to it from both of these if/else blocks, so totally makes sense.

trisapeace (commented):
Hi @tmackinnon . Whew! This is a tricky one isn't it. I think you're doing a great job figuring it out.

I didn't look at the testing file since some of it might not be relevant after changes based on my other comments. Take a look and let me know if you have any questions!

server/db/db.js
```diff
@@ -1295,6 +1295,34 @@ async function close() {
...
 
+async function getInactiveGatewaysWithClient(client, pgClient) {
+  try {
+    const results = await helpers.runQuery(
+      'getInactiveGatewaysWithClient',
+      `
+      SELECT *
+      FROM gateways g
+      WHERE g.client_id = $1
+      AND g.sent_vitals_alert_at IS NULL
```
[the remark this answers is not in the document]
trisapeace: As discussed, I was wrong and we _are_ setting the `sent_vitals_alert_at` back to `NULL`. So this code should be fine.

server/db/db.js
```diff
@@ -1295,6 +1295,37 @@ async function close() {
...
+      AND g.is_sending_vitals = true
+      LIMIT 1
+      `,
+      [client.id],
+      pool,
+      pgClient,
+    )
+
+    if (results !== undefined && results.rows.length > 0) {
+      return results.rows.map(r => createGatewayFromRow(r, [client]))
```
trisapeace, line 1316: Nice. I like how you passed `client` to the mapping function to avoid unnecessary DB calls.

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
```diff
@@ -260,4 +299,151 @@ describe('vitals.js unit tests: checkButtonHeartbeat', () => {
...
+      expect(helpers.logSentry).to.be.calledThrice
+    })
+
+    it('should send the initial disconnection messages for both buttons to Sentry', async () => {
+      await vitals.checkButtonHeartbeat()
+      expect(helpers.logSentry).to.be.calledWith(
+        `Disconnection: ${this.buttonA.client.displayName} ${this.buttonA.displayName} Button delay is ${heartbeatThreshold + 1} seconds.`,
+      )
+      expect(helpers.logSentry).to.be.calledWith(
+        `Disconnection: ${this.buttonB.client.displayName} ${this.buttonB.displayName} Button delay is ${heartbeatThreshold + 1} seconds.`,
```
trisapeace, line 339: In general, it's better to have just a single `expect` per unit test. Because a test function will stop immediately when an `expect` fails, if the first one fails then it won't even try to run the second, thus hiding a piece of information that might be useful to us when looking at the test results.

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
```diff
@@ -260,4 +299,151 @@ describe('vitals.js unit tests: checkButtonHeartbeat', () => {
...
+      )
+    })
+
+    it('should not send a reconnection message to Sentry', async () => {
+      await vitals.checkButtonHeartbeat()
+      expect(helpers.logSentry).to.not.be.calledWithMatch(`Reconnection:`)
+    })
```
trisapeace, line 346: I wonder whether you need this check since you also have one making sure that `helpers.logSentry` is called exactly three times.

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
```diff
@@ -260,4 +299,151 @@ describe('vitals.js unit tests: checkButtonHeartbeat', () => {
...
+
+    it('should send one reconnection message to Sentry', async () => {
+      await vitals.checkButtonHeartbeat()
+      expect(helpers.logSentry).to.be.calledWith(`Reconnection: ${this.buttonB.client.displayName} ${this.buttonB.displayName} Button.`)
+    })
+
+    it('should send the client message to Sentry', async () => {
+      await vitals.checkButtonHeartbeat()
+      expect(helpers.logSentry).to.be.calledWith(
+        `Button status change for: ${this.buttonA.client.displayName}. The following buttons have been disconnected: ${this.buttonA.displayName}. The following buttons have been reconnected: ${this.buttonB.displayName}.`,
```
trisapeace, line 413: Cool. I like how this message looks.

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
```diff
@@ -41,6 +44,7 @@ describe('vitals.js unit tests: checkButtonHeartbeat', () => {
     sandbox.spy(helpers, 'log')
 
     this.sendNotificationStub = sandbox.stub()
+    vitals.__set__('sendNotification', this.sendNotificationStub)
```
trisapeace, line 47: I wonder if you need this. Is `checkButtonHeartbeat` calling `sendNotification` at all?

trisapeace (commented):
Tests look good!

server/vitals.js
```diff
@@ -161,7 +185,20 @@ async function checkButtonHeartbeat() {
...
+            const gateways = await db.getInactiveGatewaysWithClient(client)
+            if (gateways.length > 0) {
+              // Store the client info
+              if (!clientMessages[client.id]) {
+                clientMessages[client.id] = {
+                  client,
+                  disconnectedButtons: [],
+                  reconnectedButtons: [],
+                }
+              }
```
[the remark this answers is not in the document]
trisapeace: As we talked about on a call, since the `sent_vitals_alert_at` _is_ in fact being set to `NULL` when the gateway comes back online, this whole comment can be ignored!

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
tmackinnon: Ahh no, good catch! I originally added this when my PR was going to using the `sendNotification` function, but since I've changed it to just log a Sentry message instead, this is not needed. Thanks!

server/vitals.js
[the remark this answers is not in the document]
tmackinnon: Thanks Theresa! :)

server/vitals.js
tmackinnon: Thanks @trisapeace! The button order will be based on the `buttonVitals` array, which comes from a db query that orders the buttons based on the `created_at` date in the `buttons_vitals_cache` table.

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
tmackinnon: Good point! I'll split these two up.

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
tmackinnon: That's a good point, this case seems a bit redundant so I'll remove it. Thanks!

server/vitals.js
trisapeace: What do you think about making it alphabetical by button display name?
I think you can do that before the `join` like this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

tmackinnon: Oh great idea - added!

server/vitals.js
```diff
@@ -140,8 +140,34 @@ async function checkButtonBatteries() {
   }
 }
 
+function logClientMessage(clientMessages) {
```
micahdbak, line 143: This is a really cool solution to use a JavaScript object as a map from client ID's to information about disconnected / reconnected buttons. However, I think without knowing that that is how you've implemented this, someone would probably be confused. I think it would be worthwhile to add a couple comments above `logClientMessage` that explains how `clientMessages` is used in the function, and how it should be created. I.e., explain that `clientMessages[clientId]` gives information for the client represented by `clientId`.

On another note, I think the variable name `clientMessages` is a little vague... Maybe it can be replaced with something like `buttonStatussesPerClient`, or something. To quote some advice Theresa gave me: a variable name should be verbose enough that you won't need comments to understand how it will be used.

server/vitals.js
```diff
@@ -140,8 +140,34 @@ async function checkButtonBatteries() {
...
+  // Check that clientMessages contains data
+  if (Object.keys(clientMessages).length > 0) {
```
micahdbak, line 145: If `clientMessages` is empty (`{}`), then `Object.values(clientMessages)` will return an empty array: `[]`. `forEach` on an empty array simply won't iterate, so I think this outer check for if there are no keys in `clientMessages` is not needed.

server/vitals.js
```diff
@@ -159,21 +185,44 @@ async function checkButtonHeartbeat() {
...
+          clientMessages[client.id].reconnectedButtons.push(button.displayName)
           await db.updateButtonsSentVitalsAlerts(button.id, false)
-
-          // TODO Also send a text message to Responders and Heartbeat Phone Numbers once we know that these messages are reliable
         }
       }
     }
+
+    // Log one message per client
+    logClientMessage(clientMessages)
```
micahdbak, line 225: Yay! I love how the functionality of logging the messages for every client comes down to this one function and one argument. It's a really clean solution; well done!

micahdbak (commented):
Looks really good so far! I just have some comments about rewording some portions, and adding more comments / explanations. The functionality and logic is well done!

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
```diff
@@ -260,4 +296,150 @@ describe('vitals.js unit tests: checkButtonHeartbeat', () => {
       expect(db.updateButtonsSentVitalsAlerts).to.not.be.called
     })
   })
+
+  describe("when client that is sending vitals's two buttons that are sending vitals last seen exceeds the threshold and no alerts have been sent", () => {
```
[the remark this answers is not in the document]
tmackinnon: This is totally understandable, and I completely agree haha. I found the wording of the pre-existing tests really confusing but was trying my best to match it for consistency, but ultimately I was struggling. 

Essentially for this test, there is a client that is sending vitals, and this client has two buttons (that are both sending vitals) and both buttons have a 'last seen' that exceeds the threshold and no alerts have been sent for either button. So there's a lot of factors.

I wonder if I could just take out the "sending vitals" bits and have: `when the same client has two buttons where their last seen exceeds the threshold and no alerts have been sent`. Is that a bit more clear?

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
```diff
@@ -260,4 +296,150 @@ describe('vitals.js unit tests: checkButtonHeartbeat', () => {
...
+
+    it('should send the client message to Sentry, with the buttons display name in alphabetical order', async () => {
+      await vitals.checkButtonHeartbeat()
+      expect(helpers.logSentry).to.be.calledWith(
+        `Button status change for: ${this.buttonA.client.displayName}. The following buttons have been disconnected: ${this.buttonA.displayName}, ${this.buttonB.displayName}.`,
+      )
+    })
+  })
+
+  describe('when client that is sending vitals, has one button sending vitals that has just disconnected, one button sending vitals that has reconnected and one button sendings vitals that remains connected', () => {
```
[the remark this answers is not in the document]
tmackinnon: Oh this is a really good idea and would help with your other comment on the wording as well! I'll give this a try!

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
```diff
@@ -3,13 +3,16 @@ const { expect, use } = require('chai')
...
 
 // In-house dependencies
 const { helpers, factories } = require('brave-alert-lib')
 const db = require('../../../db/db')
-const { buttonFactory, buttonsVitalFactory } = require('../../testingHelpers')
-const vitals = require('../../../vitals')
+const { buttonFactory, buttonsVitalFactory, gatewayFactory } = require('../../testingHelpers')
+require('../../mocks/tMock')
+
+const vitals = rewire('../../../vitals')
```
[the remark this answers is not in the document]
tmackinnon: Ahh this was from when I originally was going to be sending a twilio message vs a Sentry log. I will remove the rewire and put back the `const vitals = require('../../../vitals')`

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
```diff
@@ -260,4 +296,150 @@ describe('vitals.js unit tests: checkButtonHeartbeat', () => {
...
+        createdAt: exceededThresholdTimestamp,
+        button: this.buttonB,
+      })
+      sandbox.stub(db, 'getRecentButtonsVitals').returns([this.buttonsVitalB, this.buttonsVitalA])
+      sandbox.stub(db, 'getDisconnectedGatewaysWithClient').returns([])
+    })
+
+    it('should log three Sentry messages', async () => {
+      await vitals.checkButtonHeartbeat()
+      expect(helpers.logSentry).to.be.calledThrice
```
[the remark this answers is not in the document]
tmackinnon: Thanks for asking about this Micah! The reason I had this in there is because I want to ensure that there's exactly 3 messages being logged to Sentry and no more than that. So this test case would fail if a 4th Sentry message was logged. This is important to test because we only want to log one "client message" per client, even when a client has multiple buttons disconnected. 

If we took this test case out and my code wasn't performing as expected - it could potentially log two "client messages" since there are two buttons that have disconnected. My other test cases would still pass, but I wouldn't know that I was logging multiple "client messages" (when I was only expecting to log one).

server/vitals.js
tmackinnon: This is great feedback! I could definitely add some clarity here and change the variable name (I always struggle with these).

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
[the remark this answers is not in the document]
micahdbak: Ah, makes sense! Thanks for explaining!

server/vitals.js
tmackinnon: This shouldn't be an issue because if `clientMessages` is empty then it wouldn't even hit the line with `Object.values(clientMessages)` because doesn't pass the conditional `if (Object.keys(clientMessages).length > 0)`. So the function will just be called and nothing will happen.

server/vitals.js
tmackinnon: Thanks! :)

server/vitals.js
micahdbak: It's not an issue; I'm just saying that the outer check for `Object.keys(clientMessages).length > 0` is unnecessary, as the inner `forEach` will do nothing if `clientMessages` is empty.

micahdbak: So the function can be simplified :)

tmackinnon: Ahh I misunderstood but I see what you mean now! That's a very good point, I'll take out the conditional statement!

trisapeace (commented):
Sorry @tmackinnon , I didn't get a chance to finish looking at this today. I'll just submit these couple of reminder comments anyways.

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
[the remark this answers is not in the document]
tmackinnon: Just to follow-up on this - I re-organized the tests with nested describes as per your other comment below. Would love your feedback on that. :)

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
```diff
@@ -39,225 +41,410 @@ describe('vitals.js unit tests: checkButtonHeartbeat', () => {
...
-      await vitals.checkButtonHeartbeat()
-      expect(db.updateButtonsSentVitalsAlerts).to.not.be.called
+      it('should not send a client message for the disconnected button', async () => {
+        expect(helpers.logSentry).to.not.be.calledWith(
+          `Button status change for: ${this.button.client.displayName}. The following buttons have been disconnected: ${this.button.displayName}.`,
+        )
+      })
     })
   })
 })
```
micahdbak, line 450: There are so many test cases here; really good job being thorough! I also like how they are organized, thanks for that. :)

server/vitals.js
```diff
@@ -140,8 +140,36 @@ async function checkButtonBatteries() {
   }
 }
 
+/* 
+logClientMessage expects an object containing key-values that represent clients with recently disconnected and/or reconnected button(s)
+Each key is a clientID that holds an object with the client information and arrays of any disconnected and/or reconnected buttons
+The function will compile all this information into one message per client, and log the message to Sentry
+*/
```
micahdbak, line 147: Good explanation here!

server/vitals.js
```diff
@@ -140,8 +140,36 @@ async function checkButtonBatteries() {
...
+function logClientMessage(clientButtonStatusChanges) {
+  // Loop through each client to create the log message
+  Object.values(clientButtonStatusChanges).forEach(({ client, disconnectedButtons, reconnectedButtons }) => {
```
micahdbak, line 150: Super convenient that you can do this.

server/test/unit/vitalsTest/checkButtonHeartbeatTest.js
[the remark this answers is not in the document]
micahdbak: Was really well done! Thanks for doing that!

server/db/db.js
```diff
@@ -1295,6 +1295,36 @@ async function close() {
   }
```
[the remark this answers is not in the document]
trisapeace: Thanks!

trisapeace (approved):
This is great. Good job, @tmackinnon !

PR #209, opened 2023-12-01
Reworked the message clients API call to be a PA API call (CU-w9bcb5)

micahdbak:
### Added

- Integration tests for db.getActiveClients (CU-w9bcb5).
- Unit tests for pa.handleMessageClients (CU-w9bcb5).

### Changed

- API call /api/message-clients to become the PA API call /pa/message-clients, now authenticating with Google (CU-w9bcb5).

## Summary

The previous implementation of this API call was under /api/message-client, and didn't authenticate with the same measures that other PA API calls use.

As this API call will only be used from PA, this PR moves this API call into pa.js, and authenticates using googleHelpers.paAuthorize.

It also implements new testing for the created DB function, and handle function.

## Example Usage

POST https://chatbot(-dev).brave.coop/pa/message-clients
(body; JSON):
```json
{
    "googleIdToken": "(Google ID token as created from PA)",
    "twilioMessage": "This is a test text message sent to the active clients in the Brave servers."
}
```

(successful response; JSON):
```json
{
    "status": "success",
    "twilioMessage": "This is a test text message sent to the active clients in the Brave servers.",
    "successfullyMessaged": [
        {
            "to": "++17781111111",
            "from": "+16040000000",
            "clientId": "(client ID as stored in DB)",
            "clientDisplayName": "Example Client 1"
        },
        {
            "to": "++17782222222",
            "from": "+16040000000",
            "clientId": "(client ID as stored in DB)",
            "clientDisplayName": "Example Client 2"
        },
        {
            "to": "++17783333333",
            "from": "+16040000000",
            "clientId": "(client ID as stored in DB)",
            "clientDisplayName": "Example Client 3"
        }
    ],
    "failedToMessage": [
        {
            "to": "++17784444444",
            "from": "+16041111111",
            "clientId": "(client ID as stored in DB)",
            "clientDisplayName": "Example Client 4"
        },
    ]
}
```

## New Test Cases

- [x] db.js integration tests: getActiveClients
  - [x] if there are no clients
    - [x] should return an empty array
  - [x] if there are only active clients
    - [x] should return all of and only the active clients
  - [x] if there are only inactive clients
    - [x] should return an empty array
    - [x] should not return any of the inactive clients
  - [x] if there are active and inactive clients
    - [x] should return all of and only the active clients
    - [x] should not return any of the inactive clients
- [x] pa.js unit tests: handleMessageClients
  - [x] for a request when Twilio is operating as expected
    - [x] should respond with status 200
    - [x] should respond with the Twilio message as twilioMessage
    - [x] should respond with the expected phone numbers in successfullyMessaged
    - [x] should respond with the expected phone numbers in failedToMessage
    - [x] should call twilioHelpers.sendTwilioMessage for all active clients
  - [x] for a request when Twilio is not operating as expected
    - [x] should respond with status 200
    - [x] should respond with the Twilio message as twilioMessage
    - [x] should respond with no phone numbers in successfullyMessaged
    - [x] should respond with all phone numbers in failedToMessage
    - [x] should call twilioHelpers.sendTwilioMessage for all active clients

## Test Plan

- [x] existing test cases pass
- [x] interact with /pa/message-clients on the dev server using postman
  - [x] Submit valid Google ID Token & Twilio text message to /pa/message-clients while this PR is deployed to dev; observe that active clients in the dev database are texted
  - [x] Ran SQL query for db.getActiveClients within the dev database SQL prompt to ensure it works as expected


trisapeace:
Task linked: [CU-w9bcb5 Add a way to send a custom message to all active responder and system status phones in Buttons and Sensors](https://app.clickup.com/t/w9bcb5)

server/db/db.js
```diff
@@ -148,15 +148,15 @@ async function getActiveClients(pgClient) {
...
-	ORDER BY c.display_name;
+        FROM clients c
+        INNER JOIN (
+          SELECT DISTINCT client_id AS id
+          FROM buttons
+          WHERE is_sending_alerts AND is_sending_vitals
+        ) AS b
+        ON c.id = b.id
+        WHERE c.is_sending_alerts AND c.is_sending_vitals
+        ORDER BY c.display_name;
```
ahmaddaadaa, line 159: I don't see any differences in this part of the code. could you explain what has changed?

micahdbak: Hi Ahmad, the previous code had these lines indented with tab characters (unlike the rest of the codebase), but I've changed these lines to be indented with spaces (*like* the rest of the codebase). No logic has changed; just indentation :) .

server/test/unit/paTest/handleMessageClientsTest.js
```diff
@@ -0,0 +1,268 @@
...
+  {
+    id: 'client-duplicate-and-multiple-phone-numbers',
+    displayName: 'Client Duplicate And Multiple Phone Numbers',
+    responderPhoneNumbers: [successfulPhoneNumbers[7]],
+    fallbackPhoneNumbers: [successfulPhoneNumbers[8], successfulPhoneNumbers[9]],
+    heartbeatPhoneNumbers: [successfulPhoneNumbers[7]],
+    fromPhoneNumber: testFromPhoneNumber,
+  },
+  {
+    id: 'client-bad-from-phone-number',
```
trisapeace, line 95: I like that your id is used to describe the test case.

server/test/unit/paTest/handleMessageClientsTest.js
```diff
@@ -0,0 +1,268 @@
...
+// handleMessageClients needs:
+// - id
+// - displayName
+// - responderPhoneNumbers
+// - fallbackPhoneNumbers
+// - heartbeatPhoneNumbers
+// - fromPhoneNumber
+// from each client
+const activeClients = [
+  {
```
trisapeace, line 46: To better emulate what is returned from `db.getActive Clients` you coule use `factories.clientFactory` in `brave-alert-lib` to make `client` objects, if you want.

server/test/unit/paTest/handleMessageClientsTest.js
```diff
@@ -0,0 +1,268 @@
...
+      activeClients.forEach(client => {
+        // gather all phone numbers for this client
+        const phoneNumbers = []
+        phoneNumbers.push(...client.responderPhoneNumbers, ...client.fallbackPhoneNumbers, ...client.heartbeatPhoneNumbers)
+
+        // gather all unique phone numbers for this client
+        const uniquePhoneNumbers = new Set()
+        phoneNumbers.forEach(phoneNumber => {
+          uniquePhoneNumbers.add(phoneNumber)
+        })
```
trisapeace, line 180: I see that you've used the same code in your test as in the production code that you're testing. So if there is a bug in this implementation, you won't find it. Can you create `expectedArguments` in a different way here?

server/test/unit/paTest/handleMessageClientsTest.js
```diff
@@ -0,0 +1,268 @@
...
+            toPhoneNumber: phoneNumber,
+            fromPhoneNumber: client.fromPhoneNumber,
+            twilioMessage: testTwilioMessage,
+          })
+        })
+      })
+
+      // expect twilioHelpers.sendTwilioMessage to be called with each of the expected arguments
+      expectedArguments.forEach(args => {
+        expect(twilioHelpers.sendTwilioMessage).to.be.calledWith(args.toPhoneNumber, args.fromPhoneNumber, args.twilioMessage)
```
trisapeace, line 194: I think that this `forEach` loop has that multiple-expects-in-the-same-test issue.

What you could try to do is try doing the multiple comparisons on the mock's calls (using `sinon.getCall(n)` and calling `.args` on the results) and then use just a single `expect` to verify the whole lot. 

Might get a little ugly though. So maybe only worth it if you think there is value in knowing exactly which of the arguments sent and which ones didn't.

trisapeace (commented):
Looks good. Mostly just suggestions around the test code. Thanks for being so thorough!

server/pa.js
```diff
@@ -97,9 +98,74 @@ async function handleAwsDeviceRegistration(req, res) {
...
+      for (const client of clients) {
+        // create array of all phone numbers for this client
+        const phoneNumbers = []
+        phoneNumbers.push(...client.responderPhoneNumbers, ...client.fallbackPhoneNumbers, ...client.heartbeatPhoneNumbers)
+
+        // create set of all unique phone numbers for this client
+        const uniquePhoneNumbers = new Set()
+        phoneNumbers.forEach(phoneNumber => {
+          uniquePhoneNumbers.add(phoneNumber)
+        })
```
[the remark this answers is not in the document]
micahdbak: Weird, I remember there being a reason why I split this into two lines, but I just tested your recommendation on Node and it worked perfectly fine. I'll try this out! :)

server/test/integration/dbTest/getActiveClientsTest.js
```diff
@@ -0,0 +1,255 @@
...
+
+    clients.push(client)
+  }
+
+  return clients
+}
+
+// returns an array of client objects that are deemed inactive
+async function dbInsertInactiveClients() {
+  const inactiveClientsOptions = [
```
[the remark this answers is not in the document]
micahdbak: Luckily, when these test cases fail, chai shows the expected output as well as received output, so the tester can read over those and see what's different. I think that's sufficient for debugging... what do you think?

server/test/unit/paTest/handleMessageClientsTest.js
micahdbak: That is a good point, but my main worry here is that it'll make the file unnecessarily long. I'd probably iterate over the existing activeClients array using 'forEach', then create a new array with the outputs from factories.clientFactory. But this would make no difference, as the only values used by handleMessageClients are those explicitly set in activeClients, so I feel like it's unnecessary overhead. What do you think?

server/test/unit/paTest/handleMessageClientsTest.js
micahdbak: I can't make it different, unfortunately, unless this code was exported as a helper function from pa.js, or in brave-alert-lib. But then I feel I ought to have test cases for that function, too, and so on... :rofl:

CHANGELOG.md
```diff
@@ -11,16 +11,19 @@ the code was deployed.
 
 ## [unreleased]
 
+### Added
+
+- Translations for the Alert Type in the JSON file for both English and Spanish (CU-86795d2nb).
+- Integration tests for db.getActiveClients (CU-w9bcb5).
+- Unit tests for pa.handleMessageClients (CU-w9bcb5).
```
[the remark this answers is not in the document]
trisapeace: Sure. Sounds good to me. Thanks!

server/test/integration/dbTest/getActiveClientsTest.js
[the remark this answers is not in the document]
trisapeace: Fantastic. Thanks for checking for me.

server/test/unit/paTest/handleMessageClientsTest.js
micahdbak: Sounds like it would get ugly. but I'm willing to give this a shot, since I don't like the way it's implemented. Could you elaborate on how I would use `sinon.getCall(n)` in the mocks?

server/test/unit/paTest/handleMessageClientsTest.js
trisapeace: Mmm, maybe I'm missing something. It does add more overhead in terms of what the code is doing and the space it takes up in memory, but I don't think it significantly adds to the length of the file. In this array that you're already making, where you have something like 
```
const activeClients = [
  {
    id: 'client-just-responder-phone-number',
    displayName: 'Client Just Responder Phone Number',
    responderPhoneNumbers: [successfulPhoneNumbers[0]],
    fallbackPhoneNumbers: [],
    heartbeatPhoneNumbers: [],
    fromPhoneNumber: testFromPhoneNumber,
  },
  ...
]
```

instead you'd have

```
const activeClients = [
   factories.clientFactory({
    id: 'client-just-responder-phone-number',
    displayName: 'Client Just Responder Phone Number',
    responderPhoneNumbers: [successfulPhoneNumbers[0]],
    fallbackPhoneNumbers: [],
    heartbeatPhoneNumbers: [],
    fromPhoneNumber: testFromPhoneNumber,
  }),
  ...
]
```

One advantage it would give you is that if the client object changes for some reason in a way that is relevant to the code, then you may get those changes for free by using the factory. It will also make it easier to identify where all the client objects are being used if you do need to go through and make manual changes to all of them.

server/test/unit/paTest/handleMessageClientsTest.js
trisapeace: Could you hardcode it, considering you know what clients and phone numbers you've used?

server/test/unit/paTest/handleMessageClientsTest.js
micahdbak: Ah, that's a good point, that isn't so bad

server/test/unit/paTest/handleMessageClientsTest.js
micahdbak: True, I could! I can add this at the top of the file, then. :)

server/test/unit/paTest/handleMessageClientsTest.js
trisapeace: Haha, ok.

So when you have a mock like `twilioHelpers.sendTwilioMessage`, you can use `.getCall(n)` to learn about the nth time it was called. So the first time, would be `twilioHelpers.sendTwilioMessage.getCall(0)` and you can get what arguments were passed into by calling `.args` on the results. So to check the array of arguments sent to the mock, you could do `twilioHelpers.sendTwilioMessage.getCall().args` and compare each of those arguments to what you would expect.

I thought that there was a way to get an array of all the calls using something like `twilioHelpers.sendTwiliomessage.calls` but I don't see it in the [documentation](https://sinonjs.org/releases/v17/spy-call/) anymore, which is too bad because then you'd be able to loop over that to get all the times it was called.

But now that I'm writing this out, I think that it would enforce an order to the calls. But in reality, we don't care if they are called in a particular order, just that they are all called exactly one time. Hmmmmm.

micahdbak: That would enforce the order of the calls... In doing it this way, though, are you suggesting that I make several test cases for each call, or to do all this in one?

trisapeace: I think I've come around to just leaving your test as it was to begin with. I think doing it with only a single `expect` will end up being difficult to read the output and difficult to code. So I take back my initial suggestion, but at least now you know that the `.getCall` and `args` features exist in case they come in handy elsewhere in the future.

server/test/unit/paTest/handleMessageClientsTest.js
```diff
@@ -139,24 +148,22 @@ describe('pa.js unit tests: handleMessageClients', () => {
       )
     })
 
-    it('should respond with status 200', () => {
+    it.only('should respond with status 200', () => {
```
trisapeace, line 151: Just a remind to take out the `only`s before merging :smile:

server/test/unit/paTest/handleMessageClientsTest.js
micahdbak: Yes! That feature does sound like it would come in handy in the future, so I'll keep that in my back pocket. Thanks for exploring this with me!

server/test/unit/paTest/handleMessageClientsTest.js
micahdbak: Done!

server/test/unit/paTest/handleMessageClientsTest.js
micahdbak: Done!

server/test/integration/dbTest/getActiveClientsTest.js
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  return clients
+}
+
+describe('db.js integration tests: getActiveClients', () => {
+  beforeEach(async () => {
+    // ensure database is empty before starting each case
+    await db.clearTables()
+  })
```
tmackinnon, line 198: Looks good! All cases seem to be covered in these tests.

server/test/unit/paTest/handleMessageClientsTest.js
```diff
@@ -0,0 +1,226 @@
...
+    fromPhoneNumber: testFromPhoneNumber,
+  }),
+  factories.clientFactory({
+    id: 'client-bad-from-phone-number',
+    displayName: 'Client Bad From Phone Number',
+    responderPhoneNumbers: [failPhoneNumbers[0]],
+    fallbackPhoneNumbers: [],
+    heartbeatPhoneNumbers: [],
+    fromPhoneNumber: testBadFromPhoneNumber,
+  }),
```
tmackinnon, line 110: I'm glad you included a client with a phone number that fails, among the successful active clients.

server/test/unit/paTest/handleMessageClientsTest.js
```diff
@@ -0,0 +1,226 @@
...
+
+    it('should respond with all phone numbers in failedToMessage', () => {
+      // create array of all phone numbers (successful and fail)
+      const allPhoneNumbers = [...successfulPhoneNumbers, ...failPhoneNumbers]
+
+      // create array of to phone numbers
+      const toPhoneNumbers = this.res.body.failedToMessage.map(twilioTraceObject => twilioTraceObject.to)
+
+      // expect all (failed) to phone numbers to be equal to all phone numbers
+      expect(toPhoneNumbers).to.eql(allPhoneNumbers)
```
tmackinnon, line 216: I appreciate your comments in this, this makes understanding your logic so much easier!

tmackinnon (approved):
This looks great Micah, well done! Since most of this code mirrors what's in BraveSensor, I don't have any questions/concerns. Approved. :)

PR #213, opened 2024-01-05
Disable and reject the reset chatbot flow if it occurs (CU-860r8k57h).

micahdbak:
This PR disables and rejects the reset chatbot flow. To do this, it has the `getClientMessageForResetMessage` argument to `BraveAlerter` return `null`, which disables the alert state machine transition from `STARTED` or `WAITING_FOR_REPLY` to `RESET`. The result is no change to the existing Buttons chatbot.

### Changed

- BraveAlerterConfigurator to reject the RESET state (CU-860r8k57h).

## Test plan

- [x] Existing test cases pass
- [x] Deploy to dev server
  - [x] Interaction with chatbot works as expected
    - [x] Initiate session with RAK button press, select incident category, end session
    - [x] The reset state is not entered, ever

### DIFF
--- a/server/BraveAlerterConfigurator.js
+++ b/server/BraveAlerterConfigurator.js
@@ -110,7 +110,7 @@ class BraveAlerterConfigurator {
             return {
               respondedByPhoneNumber: session.respondedByPhoneNumber,
               replacementReturnMessageToRespondedByPhoneNumber: null,
-              replacementReturnMessageToOtherResponderPhoneNumbers: null // Don't send a message to the other responder phone numbers
+              replacementReturnMessageToOtherResponderPhoneNumbers: null,
             }
           }
 
--- a/server/BraveAlerterConfigurator.js
+++ b/server/BraveAlerterConfigurator.js
@@ -100,21 +100,14 @@ class BraveAlerterConfigurator {
 
         // If the SMS came from the session's respondedByPhoneNumber
         if (alertSession.respondedByPhoneNumber === session.respondedByPhoneNumber) {
-          // If the client sends a request to reset, reject it by not updating the session as stored in the database
-          // NOTE: this *should* never happen, as the getClientMessageForRequestToReset function returns null
-          // Thus, the alert state machine should never transition from `STARTED` or `WAITING_FOR_REPLY` to `RESET`
-          if (alertSession.alertState === CHATBOT_STATE.RESET) {
-            await db.rollbackTransaction(pgClient)
-
-            // Don't send messages to anyone; an error on part of Brave or Twilio likely caused the `RESET` state to be entered
-            return {
-              respondedByPhoneNumber: session.respondedByPhoneNumber,
-              replacementReturnMessageToRespondedByPhoneNumber: null,
-              replacementReturnMessageToOtherResponderPhoneNumbers: null,
-            }
-          }
-
           if (alertSession.alertState) {
+            // If the client sends a request to reset, reject it by throwing an Error. This will result in the session not updating in the database.
+            // NOTE: this *should* never happen, as the getClientMessageForRequestToReset function returns null. Thus, the alert state machine should
+            // never transition from `STARTED` or `WAITING_FOR_REPLY` to `RESET`.
+            if (alertSession.alertState === CHATBOT_STATE.RESET) {
+              throw new Error('Alert state was RESET; this should not happen for Buttons.')
+            }
+
             session.chatbotState = alertSession.alertState
           }
 
@@ -153,7 +146,7 @@ class BraveAlerterConfigurator {
       case CHATBOT_STATE.STARTED:
       case CHATBOT_STATE.WAITING_FOR_REPLY:
         if (toAlertState === CHATBOT_STATE.RESET) {
-          returnMessage = null
+          returnMessage = t('thankYou', { lng: language })
         } else {
           returnMessage = t('incidentCategoryRequest', {
             lng: language,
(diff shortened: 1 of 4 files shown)


trisapeace:
Task linked: [CU-860r8k57h Let clients reset their Sensor through the chatbot given >N alerts have been sent during this session](https://app.clickup.com/t/860r8k57h)

server/BraveAlerterConfigurator.js
```diff
@@ -196,6 +220,11 @@ class BraveAlerterConfigurator {
     return returnMessage
   }
 
+  getClientMessageForRequestToReset() {
+    // disable reset functionality of the chatbot by returning null
```
trisapeace, line 224: Thanks for putting this comment.

server/BraveAlerterConfigurator.js
```diff
@@ -99,6 +100,20 @@ class BraveAlerterConfigurator {
 
         // If the SMS came from the session's respondedByPhoneNumber
         if (alertSession.respondedByPhoneNumber === session.respondedByPhoneNumber) {
+          // If the client sends a request to reset, reject it by not updating the session as stored in the database
+          // NOTE: this *should* never happen, as the getClientMessageForRequestToReset function returns null
+          // Thus, the alert state machine should never transition from `STARTED` or `WAITING_FOR_REPLY` to `RESET`
+          if (alertSession.alertState === CHATBOT_STATE.RESET) {
```
trisapeace, line 106: What do you think about doing this check before even starting the transaction?

And since this shouldn't ever happen, what do you think about adding a `logError` message if it does happen?

server/BraveAlerterConfigurator.js
```diff
@@ -99,6 +100,20 @@ class BraveAlerterConfigurator {
...
+            await db.rollbackTransaction(pgClient)
+
+            // Don't send messages to anyone; an error on part of Brave or Twilio likely caused the `RESET` state to be entered
```
trisapeace, line 109: Hmmm, interesting situation. 

I'm not sure that you do want to return anything special in this case. If you were to just `return { respondedByPhoneNumber: session.respondedByPhoneNumber }` at the bottom of the function like all the other paths, would it send the "Thank you" text message? Maybe that's OK. It's what we do if there were any other errors, right?

server/BraveAlerterConfigurator.js
trisapeace: Or, if you did do it after the beginning of the transaction, you could throw an error and then it will be caught and handled like other errors.

server/BraveAlerterConfigurator.js
```diff
@@ -137,10 +152,14 @@ class BraveAlerterConfigurator {
     switch (fromAlertState) {
       case CHATBOT_STATE.STARTED:
       case CHATBOT_STATE.WAITING_FOR_REPLY:
-        returnMessage = t('incidentCategoryRequest', {
-          lng: language,
-          incidentCategories: this.createResponseStringFromIncidentCategories(incidentCategories, language),
-        })
+        if (toAlertState === CHATBOT_STATE.RESET) {
+          returnMessage = null
```
trisapeace, line 156: To follow the general chatbot principle of always sending a reply, maybe we just want to send the "Thank you" text message and `logError` the fact that we're in a state that we shouldn't ever be in?

server/BraveAlerterConfigurator.js
```diff
@@ -155,6 +174,10 @@ class BraveAlerterConfigurator {
         returnMessage = t('thankYou', { lng: language })
         break
 
+      case CHATBOT_STATE.RESET:
+        returnMessage = null
```
trisapeace, line 178: To follow the general chatbot principle of always sending a reply, maybe we just want to send the "Thank you" text message and `logError` the fact that we're in a state that we shouldn't ever be in?

server/BraveAlerterConfigurator.js
```diff
@@ -185,6 +208,7 @@ class BraveAlerterConfigurator {
         break
 
       case CHATBOT_STATE.COMPLETED:
+      case CHATBOT_STATE.RESET:
         returnMessage = null
```
trisapeace, line 212: This one seems OK as `null` because the OtherResponderPhones probably wouldn't have been expecting a text anyways.

trisapeace (commented):
Cool. Nice idea to turn off the RESET functionality entirely for Buttons.
I've just made some little comments.

server/BraveAlerterConfigurator.js
micahdbak: The main reason I'm implementing it in this way is to have the replacement messages set to `null`... Though if an Error is thrown it wouldn't have much of a difference. I'll see about additing it before the transaction, and i'll use `logError`!

server/BraveAlerterConfigurator.js
micahdbak: Actually, come to think of it, the extra return values here are extraneous as the getReturnMessageToX returns null for any RESET state transition. Do you think it should have actual messages?

server/BraveAlerterConfigurator.js
micahdbak: Ah, this answers my above comment. I agree with you here.

micahdbak: I've made it such that if the chatbot ever enters RESET (either X --> RESET, or RESET --> RESET), the chatbot simply replies 'Thank you.' This should be enough to acknowledge the client's messages without seeming broken. (This shouldn't ever happen anyways, as `getClientMessageForRequestToReset` returns null, but good to have a backup!)

PR #214, opened 2024-01-19
Rename session column 'num_button_presses' to 'number_of_alerts' (CU-86791yyvg).

micahdbak:
This is a step in the direction of merging the BraveButtons and BraveSensor databases. By moving the Session model to brave-alert-lib, both BraveButtons and BraveSensor will refer to and use the same data for sessions. For BraveButtons, this involves renaming the `num_button_presses` column to `number_of_alerts`, to match BraveSensor.

Test plan:
- [x] Existing test cases pass
- [x] Deployment to dev is successful
  - [x] Database migration script renames `num_button_presses` to `number_of_alerts`
- [x] Chatbot works as expected
  - [x] Text messages are sent for a button press
  - [x] Number of alerts column increments for every button press
  - [x] Number of presses on dashboard and over text increments correctly (from `number_of_alerts` column)
  - [x] An urgent alert is sent...
    - [x] after the 2nd press
    - [x] after every multiple of 5 presses
    - [x] after any number if it's been more than 2 minutes since the previous press


trisapeace:
Task linked: [CU-86791yyvg Combine Buttons and Sensors databases](https://app.clickup.com/t/86791yyvg)

server/buttonAlerts.js
```diff
@@ -13,12 +13,12 @@ function setup(braveAlerterObj) {
   braveAlerter = braveAlerterObj
 }
 
-async function handleValidRequest(button, numButtonPresses) {
+async function handleValidRequest(button, numberOfAlerts) {
```
trisapeace, line 16: This one might actually be worth leaving as `numButtonPresses` or taking out entirely. 

This function is called from `rak.js:handleButtonPress` each time it's called. Nowadays, it is always given the value of one. In the past, when we had Flic Buttons, they had the ability to send us a double press and we would count that as two presses, but it would still have generated just a single alert.

server/buttonAlerts.js
```diff
@@ -13,12 +13,12 @@ function setup(braveAlerterObj) {
...
   // Log the request
   helpers.log(
     `id: ${button.id.toString()} SerialNumber: ${
       button.buttonSerialNumber
-    } Unit: ${button.displayName.toString()} Presses: ${numButtonPresses.toString()} Is Sending Alerts?: ${
+    } Unit: ${button.displayName.toString()} Presses: ${numberOfAlerts.toString()} Is Sending Alerts?: ${
```
trisapeace, line 21: Similarly, `numButtonPresses` is always 1 here now that we're using RAK buttons. So you can probably take this out of the log entirely.

server/buttonAlerts.js
```diff
@@ -41,15 +41,15 @@ async function handleValidRequest(button, numButtonPresses) {
     const currentTime = await db.getCurrentTime(pgClient)
 
     if (currentSession === null || currentTime - currentSession.updatedAt >= helpers.getEnvVar('SESSION_RESET_TIMEOUT')) {
-      currentSession = await db.createSession(button.id, CHATBOT_STATE.STARTED, numButtonPresses, null, null, null, pgClient)
+      currentSession = await db.createSession(button.id, CHATBOT_STATE.STARTED, numberOfAlerts, null, null, null, pgClient)
```
trisapeace, line 44: Here you could hardcode the value 1.

server/buttonAlerts.js
```diff
@@ -41,15 +41,15 @@ async function handleValidRequest(button, numButtonPresses) {
...
     } else {
-      currentSession.incrementButtonPresses(numButtonPresses)
+      currentSession.incrementButtonPresses(numberOfAlerts)
```
trisapeace, line 46: Here you could either hardcode the value 1 or, if you're willing to make another change in `brave-alert-lib`, remove the parameter entirely and always increment by one.

server/buttonAlerts.js
```diff
@@ -41,15 +41,15 @@ async function handleValidRequest(button, numButtonPresses) {
...
       await db.saveSession(currentSession, pgClient)
     }
 
     await db.commitTransaction(pgClient)
 
-    if (currentSession.numButtonPresses === 1) {
+    if (currentSession.numberOfAlerts === 1) {
```
trisapeace, line 52: All the rest of the ones in this file though are good as `numberOfAlerts` because they are all coming from the `currentSession` object that was retrieved from the DB.

server/db/041-sessions-add-number-of-alerts.sql
```diff
@@ -0,0 +1,38 @@
...
+
+    -- Get the migration ID of the last file to be successfully run
+    SELECT MAX(id) INTO lastSuccessfulMigrationId
+    FROM migrations;
+
+    -- Only execute this script if its migration ID is next after the last successful migration ID
+    IF migrationId - lastSuccessfulMigrationId = 1 THEN
+    	-- Disable the set_sessions_timestamp trigger so updated_at doesn't change
+	ALTER TABLE sessions
+	DISABLE TRIGGER set_sessions_timestamp;
```
trisapeace, line 16: Thanks for remembering to disable the trigger!

server/db/db.js
```diff
@@ -493,20 +492,20 @@ async function getSessionWithSessionId(sessionId, pgClient) {
...
       VALUES ($1, $2, $3, $4, $5, $6, $7)
       RETURNING *
       `,
       [
         buttonId,
         chatbotState,
-        numButtonPresses > 1 ? ALERT_TYPE.BUTTONS_URGENT : ALERT_TYPE.BUTTONS_NOT_URGENT,
-        numButtonPresses,
+        numberOfAlerts > 1 ? ALERT_TYPE.BUTTONS_URGENT : ALERT_TYPE.BUTTONS_NOT_URGENT,
+        numberOfAlerts,
```
trisapeace, line 508: This is now always going to be sent `numberOfAlerts = 1`, so you could remove the logic `numberOfAlerts > 1 ? ALERT_TYPE.BUTTONS_URGENT : ALERT_TYPE.BUTTONS_NOT_URGENT` if you wanted. No big deal to leave it in if you want.

server/db/db.js
```diff
@@ -1014,7 +1013,7 @@ async function getDataForExport(pgClient) {
         b.display_name AS "Unit",
         b.phone_number AS "Button Phone",
         s.chatbot_state AS "Session State",
-        s.num_button_presses AS "Number of Presses",
+        s.number_of_alerts AS "Number of Presses",
```
trisapeace, line 1016: Thank you for leaving the column name the same :+1:

server/test/unit/SessionStateTest.js
trisapeace, line 1: Since the Session.js file is no longer in this repo, you can move these tests into `brave-alert-lib`.

trisapeace (commented):
Cool! Looking great. Feels very straightforward, which is a sign of clean coding.

Can you please also test whether the you get urgent button press alerts in these three cases:
- after the 2nd press,
- after every multiple of 5 presses, and
- after any if it's been more than 2 minutes since the previous press

server/buttonAlerts.js
micahdbak: Oh, interesting. Removing it is probably the best action here then; I'll do that!

server/buttonAlerts.js
micahdbak: Sure can do!

server/buttonAlerts.js
micahdbak: I'll make a small update to brave-alert-lib. :)

server/test/unit/SessionStateTest.js
micahdbak: Ah, here's tests! I was looking for some but I guess I missed these. Thanks!

server/db/db.js
micahdbak: Should this be urgent or not urgent upon creation? I'm going to start a session with 1 alert, so by the existing logic, it should be not urgent.

server/test/unit/SessionStateTest.js
micahdbak: Actually, I think I'm going to delete this test file, as it only tests the incrementButtonPresses function. This function is only used once, so I'm going to delete it, and move it's code into BraveButtons. Keeps things simpler that way; and the sessionFactory it needs is kind of device-specific.

server/db/db.js
micahdbak: I'm keeping this as ALERT_TYPE.BUTTONS_NOT_URGENT; it changes to ALERT_TYPE.BUTTONS_URGENT after another press.

micahdbak:
~ deploying to the dev server to complete the Test Plan

micahdbak:
Test Plan passes as per Theresa's request; I'm going to merge this PR.

server/db/db.js
trisapeace: Exactly. NOT_URGENT to start with :+1:

PR #215, opened 2024-01-24
Create REST API (CU-86dugzyzh)

micahdbak:
This PR is still in progress.


trisapeace:
Task linked: [CU-13kqjz8 Merge buttons and sensors backend (including brave-alert-lib)](https://app.clickup.com/t/13kqjz8)

trisapeace (commented):
We talked about this together this morning and you've made some changes that aren't reflected here yet, so I won't make any direct comments. You're on the right track and this is going to be **awesome**!!!

micahdbak:
~ Deploying to the dev server to test the new get routes

micahdbak:
~ Deploying to dev (pending passed checks) to test validation of all routes

micahdbak:
~ Deploying to the dev server

micahdbak:
~ Deploying again

server/api.js
```diff
@@ -0,0 +1,551 @@
...
+}
+
+const validateUpdateClientButton = [
+  Validator.header(['Authorization']).notEmpty(),
+  Validator.param(['clientId', 'buttonId']).notEmpty(),
+  Validator.body(['displayName', 'phoneNumber', 'buttonSerialNumber']).trim().isString().notEmpty(),
+  Validator.body(['isDisplayed', 'isSendingAlerts', 'isSendingVitals']).isBoolean(),
+]
+
+async function handleUpdateClientButton(req, res) {
```
trisapeace, line 435: Fun use case for you... what if we want to move a button from one client to another. This could happen in some of our clients who have multiple locations with different responder phone numbers who decide they want to rearrange their buttons for some reason.

The current Sensors dashboard lets users do this using the Client dropdown menu:
image

What do you think about adding a `clientId` parameter to this and the db.updateButton call?

Ditto for gateways.

server/api.js
```diff
@@ -0,0 +1,551 @@
...
+  handleGetClientGateway,
+  handleGetClientGateways,
+  handleGetClientVitals,
+  handleRegisterClient,
+  handleRegisterClientButton,
+  handleRegisterClientGateway,
+  handleUpdateClient,
+  handleUpdateClientButton,
+  handleUpdateClientGateway,
+  validateGetClient,
```
trisapeace, line 537: Thank you for keeping the `handle*` and `validate*` naming style!

server/db/db.js
```diff
@@ -24,7 +24,11 @@ pool.on('error', err => {
 types.setTypeParser(types.builtins.NUMERIC, value => parseFloat(value))
 
 function createSessionFromRow(r, allButtons) {
-  const button = allButtons.filter(b => b.id === r.button_id)[0]
+  let button // initially undefined
+
+  if (allButtons !== undefined) {
+    button = allButtons.filter(b => b.id === r.button_id)[0]
+  }
```
trisapeace, line 31: We talked about this earlier, so I'll just note down my understanding of where we got to. 

Good options are either:
* Filter out the button when you want to in the api.js functions, or
* Create new db.* functions with names that make it clear that they will not contain the full button

server/db/db.js
```diff
@@ -1190,34 +1472,43 @@ module.exports = {
...
   getRecentButtonsVitals,
   getRecentButtonsVitalsWithClientId,
   getRecentGatewaysVitals,
   getRecentGatewaysVitalsWithClientId,
   getRecentGatewaysVitalWithGatewayId,
+  getRecentSessionsWithButtonId,
   getRecentSessionsWithClientId,
   getSessionWithSessionId,
   getUnrespondedSessionWithButtonId,
-  getDisconnectedGatewaysWithClient,
```
trisapeace, line 1215: Thanks for fixing the ordering!

trisapeace (commented):
You are well on your way. This is looking fantastic. I'm glad you're having fun!

server/api.js
```diff
@@ -0,0 +1,551 @@
+/* Conventions for this API:
```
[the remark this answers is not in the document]
micahdbak: That's the hope!

server/api.js
```diff
@@ -0,0 +1,551 @@
...
+
+async function handleGetClient(req, res) {
+  const validationErrors = Validator.validationResult(req).formatWith(helpers.formatExpressValidationErrors)
+
+  try {
+    if (validationErrors.isEmpty()) {
+      const client = await db.getClientWithId(req.params.clientId)
+
+      // check that the client exists
+      if (client == null) {
```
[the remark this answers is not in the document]
micahdbak: I chose loose equals here just incase the function returns undefined. `helpers.runQuery` will return undefined upon an error, which causes a TypeError when attempting to read `results.rows`, which leads to database functions returning `null` in specific. I think there could be an off-chance that someone messes with a database function, or some extravagant error leads to something returning undefined, that I'd like to catch with this loose equals. Thoughts?

server/api.js
```diff
@@ -0,0 +1,551 @@
...
+async function handleGetClientButton(req, res) {
+  const validationErrors = Validator.validationResult(req).formatWith(helpers.formatExpressValidationErrors)
+
+  try {
+    if (validationErrors.isEmpty()) {
+      const button = await db.getButtonWithId(req.params.buttonId)
+
+      // check that this button exists and is owned by the specified client
+      // NOTE: if clientId is invalid, then the query will fail and return null
+      if (button == null || button.client.id !== req.params.clientId) {
```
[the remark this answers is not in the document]
micahdbak: All are the same! I've purposely checked loosely against null for "nully" values.

server/api.js
```diff
@@ -0,0 +1,551 @@
...
+
+async function handleGetClientButton(req, res) {
+  const validationErrors = Validator.validationResult(req).formatWith(helpers.formatExpressValidationErrors)
+
+  try {
+    if (validationErrors.isEmpty()) {
+      const button = await db.getButtonWithId(req.params.buttonId)
+
+      // check that this button exists and is owned by the specified client
+      // NOTE: if clientId is invalid, then the query will fail and return null
```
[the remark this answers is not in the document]
micahdbak: From reading the code for getButtonWithId, it should return null if clientId is invalid; i.e., the check `if (results.rows.length > 0)` fails (no rows returned), and the try-catch block completes, and then `return null` statement would run.

server/api.js
```diff
@@ -0,0 +1,551 @@
...
+const validateGetClientButton = [Validator.header(['Authorization']).notEmpty(), Validator.param(['clientId', 'buttonId']).notEmpty()]
+
+async function handleGetClientButton(req, res) {
+  const validationErrors = Validator.validationResult(req).formatWith(helpers.formatExpressValidationErrors)
+
+  try {
+    if (validationErrors.isEmpty()) {
+      const button = await db.getButtonWithId(req.params.buttonId)
+
+      // check that this button exists and is owned by the specified client
```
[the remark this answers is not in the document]
micahdbak: Interesting! I'll think about that.

server/api.js
```diff
@@ -0,0 +1,551 @@
...
+async function handleGetClientButtons(req, res) {
+  const validationErrors = Validator.validationResult(req).formatWith(helpers.formatExpressValidationErrors)
+
+  try {
+    if (validationErrors.isEmpty()) {
+      const buttons = await db.getButtonsWithClientId(req.params.clientId)
+
+      // if the query failed and returned null, the clientId is probably wrong
+      if (buttons == null) {
+        res.status(404).send({ status: 'error', message: 'Not Found' })
```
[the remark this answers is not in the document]
micahdbak: Yeah, I agree. A lot of the 404 returns would be better as either 500 or 400, so I need to go over all of those.

server/api.js
```diff
@@ -0,0 +1,551 @@
...
+const validateGetClients = Validator.header(['Authorization']).notEmpty()
+
+async function handleGetClients(req, res) {
+  const validationErrors = Validator.validationResult(req).formatWith(helpers.formatExpressValidationErrors)
+
+  try {
+    if (validationErrors.isEmpty()) {
+      const clients = await db.getClients()
+
+      res.status(200).send({ status: 'success', data: clients })
```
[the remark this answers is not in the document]
micahdbak: The validation is different for each of the non-get functions, but the surrounding bits of each handler is mostly consistent. How do you propose I factor these out? Something like this?

```javascript

async function handleApiRequest(req, res, handler) {
  const validationErrors = Validator.validationResult(req).formatWith(helpers.formatExpressValidationErrors)

  try {
    if (validationErrors.isEmpty()) {
      handler(req, res)
    } else {
      res.status(400).send({ status: 'error', message: 'Bad Request' })
      helpers.logError(`Bad request to ${req.path}: ${validationErrors.array()}`)
    }
  } catch (error) {
    res.status(500).send({ status: 'error', message: 'Internal Server Error' })
    helpers.logError(`Internal server error at ${req.path}: ${error.message}`)
  }
}

async function handleGetClient(req, res) {
  handleApiRequest(req, res, () => {
    // ... database code
  })
}

```

server/api.js
```diff
@@ -0,0 +1,551 @@
...
+  const validationErrors = Validator.validationResult(req).formatWith(helpers.formatExpressValidationErrors)
+
+  try {
+    if (validationErrors.isEmpty()) {
+      const buttonVitals = await db.getRecentButtonsVitalsWithClientId(req.params.clientId)
+      const gatewayVitals = await db.getRecentGatewaysVitalsWithClientId(req.params.clientId)
+
+      // if either of the query failed and returned null, the clientId is probably wrong
+      if (buttonVitals == null || gatewayVitals == null) {
+        res.status(404).send({ status: 'error', message: 'Not Found' })
```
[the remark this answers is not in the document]
micahdbak: It is kind of weird for an empty array to be returned if the input client ID was wrong, for example. I'll change the database function to return null in the event of an error, instead of an empty array!

server/routes.js
```diff
@@ -1,24 +1,55 @@
...
-
   app.post('/login', dashboard.submitLogin)
+
+  // new to-be-implemented API routes
+  app.get('/api/clients', api.validateGetClients, googleHelpers.paAuthorize, api.handleGetClients)
+  app.get('/api/clients/:clientId', api.validateGetClient, googleHelpers.paAuthorize, api.handleGetClient)
+  app.get('/api/clients/:clientId/buttons', api.validateGetClientButtons, googleHelpers.paAuthorize, api.handleGetClientButtons)
+  app.get('/api/clients/:clientId/buttons/:buttonId', api.validateGetClientButton, googleHelpers.paAuthorize, api.handleGetClientButton)
+  app.get(
+    '/api/clients/:clientId/buttons/:buttonId/sessions',
```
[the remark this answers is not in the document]
micahdbak: Sure! I like this.

server/api.js
```diff
@@ -0,0 +1,551 @@
...
+    helpers.logError(`Internal server error at ${req.path}: ${error.message}`)
+  }
+}
+
+const validateRegisterClient = [
+  Validator.header(['Authorization']).notEmpty(),
+  Validator.body(['displayName', 'fromPhoneNumber', 'language']).trim().isString().notEmpty(),
+  Validator.body(['responderPhoneNumbers', 'fallbackPhoneNumbers', 'heartbeatPhoneNumbers', 'incidentCategories']).isArray(),
+  Validator.body(['reminderTimeout', 'fallbackTimeout']).isInt({ min: 0 }),
+  Validator.body(['isDisplayed', 'isSendingAlerts', 'isSendingVitals']).isBoolean(),
```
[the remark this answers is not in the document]
micahdbak: Ooo, sure! I like these!

server/api.js
```diff
@@ -0,0 +1,551 @@
...
+]
+
+async function handleRegisterClient(req, res) {
```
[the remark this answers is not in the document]
micahdbak: This is to match the functions in BraveSensor. Do you think "create" is a better term to use?

server/api.js
micahdbak: Done for both of these! I agree with having the option to move buttons/gateways from one client to another.

server/db/db.js
```diff
@@ -36,27 +40,43 @@ function createClientFromRow(r) {
...
 function createButtonFromRow(r, allClients) {
-  const client = allClients.filter(c => c.id === r.client_id)[0]
+  let client // initially undefined
+
+  if (allClients !== undefined) {
+    client = allClients.filter(c => c.id === r.client_id)[0]
+  }
 
   // prettier-ignore
   return new Button(r.id, r.display_name, r.phone_number, r.created_at, r.updated_at, r.button_serial_number, r.is_displayed, r.is_sending_alerts, r.is_sending_vitals, r.sent_low_battery_alert_at, r.sent_vitals_alert_at, client)
```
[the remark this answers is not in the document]
micahdbak: Oo; agree!

server/api.js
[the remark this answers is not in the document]
micahdbak: I've done this, and put the `googleHelpers.paAuthorize` function in a new `api.authorize` function. The functions are all shorter and cleaner now :) .

server/api.js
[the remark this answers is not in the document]
micahdbak: I changed my mind on this; I'm not entirely sure where these DB functions are used, so it might be better to have the safeguard of always returning an array, and returning an empty array even if an error occurred. It makes sense logically too---if a client doesn't exist, then they can't have any buttons; therefore an empty array makes sense. I've reverted my changes to these functions and changed my new functions to follow from that.

I think I'm going to try and have all API routes return 200, 201 if OK; 400 if validation fails; and 500 if an error was thrown. I'm going to rework the existing 404s to either return an empty array, null, or throw an error.

server/api.js
```diff
@@ -0,0 +1,551 @@
...
+  const validationErrors = Validator.validationResult(req).formatWith(helpers.formatExpressValidationErrors)
+
+  try {
+    if (validationErrors.isEmpty()) {
+      const client = await db.createClient(
+        req.body.displayName,
+        req.body.responderPhoneNumbers,
```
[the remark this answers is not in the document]
micahdbak: I'm going to leave this open since I'm not sure how to use Validator to validate the individual items of a submitted Array.

I'd like to validate:
- phone numbers with regex after trim
- incident categories after trim

server/api.js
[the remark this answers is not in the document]
micahdbak: Done!

server/api.js
```diff
@@ -0,0 +1,551 @@
...
+        req.body.heartbeatPhoneNumbers,
+        req.body.incidentCategories,
+        req.body.isDisplayed,
+        req.body.isSendingAlerts,
+        req.body.isSendingVitals,
+        req.body.language,
+      )
+
+      // Should the database query fail, db.createClient should internally handle thrown errors and return either null or undefined.
+      // The status code 400 is used here as the failure was probably caused by invalid fields.
```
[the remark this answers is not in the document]
micahdbak: (Discussed under a different conversation.)

server/api.js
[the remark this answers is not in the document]
micahdbak: I've changed these to 'Create'.

server/api.js
```diff
@@ -0,0 +1,551 @@
...
+    }
+  } catch (error) {
+    res.status(500).send({ status: 'error', message: 'Internal Server Error' })
+    helpers.logError(`Internal server error at ${req.path}: ${error.message}`)
+  }
+}
+
+const validateRegisterClientGateway = [
+  Validator.header(['Authorization']).notEmpty(),
+  Validator.param(['gatewayId', 'clientId']).notEmpty(),
```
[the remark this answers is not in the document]
micahdbak: (Discussed under a different conversation.)

server/db/db.js
micahdbak: I've decided to keep the DB functions as they are, with full client objects, then filter these from within api.js.

server/api.js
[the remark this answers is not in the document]
trisapeace: Thanks for clarifying! 

I think that seeing `==` in JS code looks funny and will automatically bring up questions. Which means that if you keep it like this, then it would be great if you added a comment with your reasoning.

On the other hand, I think that writing unit tests to guard against any future dev doing funny business with the DB function would be a better safeguard than a loose equals.

I'm OK with either way!

server/api.js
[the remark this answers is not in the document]
trisapeace: Yes, that looks good to me!

In addition, if you wanted to make it easier to run the database code on its own, you could even make it a named function that you pass along to `handleApiRequest`.

server/api.js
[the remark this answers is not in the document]
trisapeace: I think you'd need to write a custom validator for it. Not sure if it's worth it. 

https://express-validator.github.io/docs/6.5.0/custom-validators-sanitizers/

server/api.js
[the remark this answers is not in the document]
trisapeace: Yes, I was thinking `Create`. I didn't realize that `Register` was used in Sensors. I like `Create` better :smile:

server/api.js
```diff
@@ -18,508 +18,368 @@
...
-      if (client == null) {
-        res.status(404).send({ status: 'error', message: 'Not Found' })
-
-        return
+// wrapper function for a given handler
+// this should preceed any handler when defining an Express route
+async function wrapper(req, res, next) {
+  // authorize with a Google ID Token submitted through the Authorization header of the request
+  // if the request is authorized, then the passed function will run
+  googleHelpers.paAuthorize(req, res, () => {
```
trisapeace, line 29: Gotcha! I don't think I understood what you meant when you said that you were moving the `paAuthorize` into the wrapper. I think you could just add it to the routes. I'll add a comment down there to explain what I mean...

server/routes.js
```diff
@@ -17,30 +17,25 @@ function configureRoutes(app) {
...
-  app.get('/api/clients/:clientId/vitals', api.validateGetClientVitals, googleHelpers.paAuthorize, api.handleGetClientVitals)
-  app.post('/api/clients', api.validateRegisterClient, googleHelpers.paAuthorize, api.handleRegisterClient)
-  app.post('/api/clients/:clientId/buttons', api.validateRegisterClientButton, googleHelpers.paAuthorize, api.handleRegisterClientButton)
-  app.post('/api/clients/:clientId/gateways', api.validateRegisterClientGateway, googleHelpers.paAuthorize, api.handleRegisterClientGateway)
-  app.put('/api/clients/:clientId', api.validateUpdateClient, googleHelpers.paAuthorize, api.handleUpdateClient)
-  app.put('/api/clients/:clientId/buttons/:buttonId', api.validateUpdateClientButton, googleHelpers.paAuthorize, api.handleUpdateClientButton)
-  app.put('/api/clients/:clientId/gateways/:gatewayId', api.validateUpdateClientGateway, googleHelpers.paAuthorize, api.handleUpdateClientGateway)
+  app.get('/api/clients', api.wrapper, api.handleGetClients)
+  app.get('/api/clients/:clientId', api.validateGetClient, api.wrapper, api.handleGetClient)
```
trisapeace, line 21: The expressJS functions can take any number of functions as parameters. So you can just add in your new wrapper function to run after `paAuthorize`. Like this: `app.get('/api/clients/:clientId', api.validateGetClient, googleHelpers.paAuthorize, api.wrapper, api.handleGetClient)`

Unless there's something that makes this not work that I don't see. Let me know if there is!

trisapeace (commented):
Gotta run to get to my volunteering. I'm still not done looking at this. I'll get back to it later this afternoon!

server/routes.js
micahdbak: My reasoning for putting into the wrapper is twofold:
a) In the future, the means of authentication may be changed to allow for clients to access routes under their client
b) With google.paAuthorize in each route, the linter requires each route to be multiline. (As there are many routes, this makes the routes file veryyy long.)

server/api.js
```diff
@@ -0,0 +1,408 @@
...
+    return
+  }
+
+  res.status(200).send({ status: 'success', data: client })
+}
+
+async function handleGetClients(req, res) {
+  const clients = await db.getClients()
+
+  res.status(200).send({ status: 'success', data: clients })
```
trisapeace, line 157: Cool! I love that these functions are so simple now!

server/db/db.js
```diff
@@ -1175,6 +1179,281 @@ async function getDisconnectedGatewaysWithClient(client, pgClient) {
...
+      LIMIT 40
+      `,
+      [buttonId],
+      pool,
+      pgClient,
+    )
+
+    if (results !== undefined && results.rows.length > 0) {
+      const allClients = await getClients(pgClient)
+      return results.rows.map(r => createSessionFromRow(r, allClients))
```
trisapeace, line 1255: Ditto here about `createSessionFromRow` expecting an array of Buttons rather than an array of Clients.

server/db/db.js
```diff
@@ -1175,6 +1179,281 @@ async function getDisconnectedGatewaysWithClient(client, pgClient) {
...
+      ],
+      pool,
+      pgClient,
+    )
+
+    if (results === undefined || results.rows.length === 0) {
+      return null
+    }
+
+    helpers.log(`Client '${displayName}' successfully updated`)
```
trisapeace, line 1423: Curious why you log this one but not others?

trisapeace (commented):
Looking good!

server/api.js
```diff
@@ -0,0 +1,409 @@
+/* Conventions for this API:
+ *  - GET method for read actions
+ *  - POST method for create actions (think POSTing a new item to a directory)
+ *  - PUT method for update actions (think PUTting an item over an existing item)
```
micahdbak, line 4: Accurate as per HTML spec

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+
+// In-house dependencies
+const { helpers, googleHelpers } = require('brave-alert-lib')
+const db = require('./db/db')
+
+// authorize function - using Google ID Tokens
+// this should be significantly reworked if providing a limited API to clients
+// NOTE: a route's validation should PRECEED the authorize function, and a route's handler should PROCEED the authorize function;
+//   e.g.: app.method('/api/thing', api.validateThing, api.authorize, api.handleThing)
+async function authorize(req, res, next) {
```
micahdbak, line 28: Rework this to NOT use Google to authenticate. Maybe just use the Brave API key?

Reason: these API routes may not be implemented from PA, if for a client facing dashboard

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+}
+
+const validateCreateClientButton = [
+  Validator.param(['clientId']).notEmpty(),
+  Validator.body(['displayName', 'phoneNumber', 'buttonSerialNumber']).trim().isString().notEmpty(),
+  Validator.body(['isDisplayed', 'isSendingAlerts', 'isSendingVitals']).trim().isBoolean(),
+]
+
+async function handleCreateClientButton(req, res) {
+  const button = await db.createButton(
```
micahdbak, line 87: rework to createDevice

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+    req.body.buttonSerialNumber,
+    req.body.isDisplayed,
+    req.body.isSendingAlerts,
+    req.body.isSendingVitals,
+    null,
+    null,
+  )
+
+  // Couldn't create button; Internal server error.
+  if (button == null) {
```
micahdbak, line 100: Comment here for nully value

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+    req.body.isSendingVitals,
+    req.body.language,
+  )
+
+  if (client == null) {
+    // will result in a status 500
+    throw new Error('Failed to create client')
+  }
+
+  res.set('Location', `${req.path}/${client.id}`) // location of newly created client
```
micahdbak, line 76: set to full path using provided things; e.g.,

`/clients/${client.id}`

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+
+async function handleGetClients(req, res) {
+  const clients = await db.getClients()
+
+  res.status(200).send({ status: 'success', data: clients })
+}
+
+const validateGetClientButton = Validator.param(['clientId', 'buttonId']).notEmpty()
+
+async function handleGetClientButton(req, res) {
```
micahdbak, line 159: rework to handleGetClientDevice

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+  const button = await db.getButtonWithId(req.params.buttonId)
```
micahdbak, line 160: rework

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+
+  // Couldn't get the button; Not found.
+  if (button == null || button.client.id !== req.params.clientId) {
```
micahdbak, line 163: is second param here necessary?

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+    return
+  }
+
+  res.status(200).send({ status: 'success', data: button })
+}
+
+const validateGetClientButtons = Validator.param(['clientId']).notEmpty()
+
+async function handleGetClientButtons(req, res) {
+  const buttons = await db.getButtonsWithClientId(req.params.clientId)
```
micahdbak, line 175: rework

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+
+  // if the query failed and returned null, the clientId is probably wrong
+  if (buttons == null) {
+    res.status(404).send({ status: 'error', message: 'Not Found' })
+
+    return
+  }
+
+  // something like this: buttons = buttons.map(button => { button.sdfs ... })
```
micahdbak, line 184: huh???

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+  // (remove single field)
+
+  res.status(200).send({ status: 'success', data: buttons })
+}
+
+const validateGetClientButtonSessions = Validator.param(['clientId', 'buttonId']).notEmpty()
+
+async function handleGetClientButtonSessions(req, res) {
+  const button = await db.getButtonWithId(req.params.buttonId)
```
micahdbak, line 193: rework

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+    return
+  }
+
+  // something like this: buttons = buttons.map(button => { button.sdfs ... })
+  // (remove single field)
+
+  res.status(200).send({ status: 'success', data: buttons })
+}
+
+const validateGetClientButtonSessions = Validator.param(['clientId', 'buttonId']).notEmpty()
```
micahdbak, line 190: rework

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+    return
+  }
+
+  res.status(200).send({ status: 'success', data: gateways })
+}
+
+const validateGetClientVitals = Validator.param(['clientId']).notEmpty()
+
+async function handleGetClientVitals(req, res) {
+  const buttonVitals = await db.getRecentButtonsVitalsWithClientId(req.params.clientId)
```
micahdbak, line 242: are these arrays?

micahdbak: also rework

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+// NOTE: clientId is submitted in the param and body of the request.
+// This is to let a button be moved from one client to another; think of the param clientId as 'from' and the body clientId as 'to'.
+const validateUpdateClientButton = [
+  Validator.param(['clientId', 'buttonId']).notEmpty(),
+  Validator.body(['clientId', 'displayName', 'phoneNumber', 'buttonSerialNumber']).trim().isString().notEmpty(),
+  Validator.body(['isDisplayed', 'isSendingAlerts', 'isSendingVitals']).trim().isBoolean(),
+]
+
+async function handleUpdateClientButton(req, res) {
+  const button = await db.getButtonWithId(req.params.buttonId)
```
micahdbak, line 310: rework

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+  // check that this button exists and is owned by the specified client
+  // NOTE: if clientId is invalid, then the query will fail and return null
+  if (button == null || button.client.id !== req.params.clientId) {
+    res.status(404).send({ status: 'error', message: 'Not Found' })
+
+    return
+  }
+
+  // attempt to update the button
+  const updatedButton = await db.updateButton(
```
micahdbak, line 321: rework

server/routes.js
```diff
@@ -1,24 +1,50 @@
...
+  app.get('/api/clients/:clientId/gateways', api.validateGetClientGateways, api.authorize, api.handleGetClientGateways)
+  app.get('/api/clients/:clientId/gateways/:gatewayId', api.validateGetClientGateway, api.authorize, api.handleGetClientGateway)
+  app.get('/api/clients/:clientId/vitals', api.validateGetClientVitals, api.authorize, api.handleGetClientVitals)
+  app.post('/api/clients', api.validateCreateClient, api.authorize, api.handleCreateClient)
+  app.post('/api/clients/:clientId/buttons', api.validateCreateClientButton, api.authorize, api.handleCreateClientButton)
+  app.post('/api/clients/:clientId/gateways', api.validateCreateClientGateway, api.authorize, api.handleCreateClientGateway)
+  app.put('/api/clients/:clientId', api.validateUpdateClient, api.authorize, api.handleUpdateClient)
+  app.put('/api/clients/:clientId/buttons/:buttonId', api.validateUpdateClientButton, api.authorize, api.handleUpdateClientButton)
+  app.put('/api/clients/:clientId/gateways/:gatewayId', api.validateUpdateClientGateway, api.authorize, api.handleUpdateClientGateway)
```
micahdbak, line 33: reorder these based on route, not http method

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+  validateGetClientButtons,
+  validateGetClientButtonSessions,
+  validateGetClientGateway,
+  validateGetClientGateways,
+  validateGetClientVitals,
+  validateUpdateClient,
+  validateUpdateClientButton,
+  validateUpdateClientGateway,
+  authorize,
+}
```
micahdbak, line 409: new route for client sessions, which displays recent sessions (maybe last 20? 40?) for a given client

server/api.js
```diff
@@ -0,0 +1,409 @@
...
+
+  // check that this button exists and is owned by the specified client
+  // NOTE: if clientId is invalid, then the query will fail and return null
+  if (button == null || button.client.id !== req.params.clientId) {
+    res.status(404).send({ status: 'error', message: 'Not Found' })
+
+    return
+  }
+
+  const sessions = await db.getRecentSessionsWithButtonId(req.params.buttonId)
```
micahdbak, line 203: rework - how many sessions does this return? Maybe add new parameter with default value, and SQL `LIMIT`?

micahdbak (commented):
General reworks include using new device DB methods instead of button DB methods, but keeping the routes button-oriented

micahdbak:
Big TODO after reworks: create integration tests for each route

trisapeace:
Task linked: [CU-86dugzyzh Create REST API for Buttons](https://app.clickup.com/t/86dugzyzh)

CCheng00 (approved):
Looks good!

